PR #437, opened 2015-04-12
Ensure all output is written before exiting (stops truncation)

mhart:
This surfaces in cases where `process.exit()` is called (from `endWork`) before all writes have flushed.

Reproducible on io.js when piping a large output to another process, eg, tail – the output ends up truncated.

Before this PR:
```console
$ cat my.css | autoprefixer -b '> 1%' | tail -3
.tap-to-enlarge {
  font-size: 1rem;
  display: bloc
```
(ie, truncation)

After this PR:
```console
$ cat my.css | autoprefixer -b '> 1%' | tail -3
  .menu-item-back > .menu-link {
    color: #fff; }

```
(no truncation)

There are a number of other places where providing a callback to `print` or even `error` should be done, but I only fixed the one where the problem is likely to occur – and that's when outputting to stdout – all other areas in the code *could* have problems, but are unlikely to (because they'll always likely output less than a buffer's worth of data at a time).

### DIFF
--- a/binary.js
+++ b/binary.js
@@ -74,10 +74,10 @@ Browsers:
   or last version: '-b "last 2 versions"'.`);
     }
 
-    // Print to stdout
-    print(str) {
+    // Print to stdout, with optional callback
+    print(str, done) {
         str = str.replace(/\n$/, '');
-        this.stdout.write(str + "\n");
+        this.stdout.write(str + "\n", done);
     }
 
     // Print to stdout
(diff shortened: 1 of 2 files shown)


mhart:
There are currently issues with two of the tests, but I suspect it's the way the tests are written – looking into it now.

mhart:
OK, tests passing now

mhart:
The other option would just be to get rid of the hard `process.exit` (except in cases of failure) and let the process flush all writes and exit gracefully. I'll look at how that affects things.

mhart:
OK, only calling `process.exit` when there's an error is probably a better way to handle this problem.

Have done this in #438 so will close this.

Issue #571, opened 2015-11-30
Bug in gradient.coffee, array out bounds


lexich:
https://github.com/postcss/autoprefixer/blob/master/lib/hacks/gradient.coffee#L78
```coffee
# Replace `to top left` to `bottom right`
  fixDirection: (params) ->
    params.splice(0, 2)
    for i in [0..params.length - 1]  #!!!!!FIX
      if params[i].type == 'div'
        break
      if params[i].type == 'word'
        params[i].value = @revertDirection(params[i].value)
```

http://coffeescript.org/#try:params%20%3D%20%5B1%2C2%2C3%5D%0Afor%20i%20in%20%5B0..params.length%5D%0A%20%20console.log(%22PRINT%22%2C%20params%5Bi%5D)

Output:
```
PRINT 1
PRINT 2
PRINT 3
PRINT undefined
```

lydell:
`for i in [0..params.length]` → `for i in [0...params.length]` is nicer than adding ` - 1` :)

lexich:
@lydell Sorry but you write the same, but this bug crush my workflow
```
ERROR in ./~/css-loader?sourceMap!./~/postcss-loader!./~/sass-loader?outputStyle=expanded&includePaths[]=%2FUsers%2Flexich%2Fcode%2Favi%2F24slash6%2Fapp%2Fstyles&includePaths[]=%2FUsers%2Flexich%2Fcode%2Favi%2F24slash6%2Fbower_components!./bower_components/card/src/scss/card.scss
Module build failed: TypeError: Cannot read property 'type' of undefined
  at Gradient.fixDirection (.../node_modules/autoprefixer/lib/hacks/gradient.js:115:22)
.....
```

lydell:
@lexich No, I've added an extra `.` ;)

lexich:
Very tricky dots, don't saw :)

lexich:
Continue in #572

ai:
Release in 6.1.2

PR #572, opened 2015-11-30
Fix: range outbound bug

lexich:
Fix issue #571

### DIFF
--- a/test/cases/gradient-fix.css
+++ b/test/cases/gradient-fix.css
@@ -1,0 +1,4 @@
+.test {
+  background-image: -webkit-linear-gradient(yellow 50%, red 50%);
+  background-image: linear-gradient(to   , red 50%);
+}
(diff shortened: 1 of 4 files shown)


TrySound:
@lexich Please, add tests where it fails

lexich:
@TrySound I add test case. Css isn't very valid, but bug is bug.

TrySound:
@lexich Squash commits please

lexich:
@TrySound OK

ai:
Release in 6.1.2

test/cases/gradient-fix.css
```diff
@@ -0,0 +1,4 @@
+.test {
+  background-image: -webkit-linear-gradient(yellow 50%, red 50%);
+  background-image: linear-gradient(to   , red 50%);
```
yisibl: Why use the wrong syntax?

> linear-gradient(to   , red 50%)

lexich:
@yisibl It's a generated code from http://bourbon.io/ and code crush in this case.

PR #598, opened 2015-12-29
Travis CI: Switch deprecated `stable` NodeJS to `5` to test the latest 5.x.x NodeJS version

ntwb:


ai:
But when nodejs will release new version, it will be automatically tested in Travis CI.

Mayb we should replace it to `latest` or something like it?

ai:
Also do you have any links that `stable` was depreacted? I am curious.

ntwb:
Yup, Travis needs to update there docs but here on the [NVM #Usage](https://github.com/creationix/nvm#usage) section of the docs its documented about the `stable` alias being deprecated.

In https://github.com/creationix/nvm/issues/870 there is a discussion about future aliases such as `LTS`, and other options...

That said though, the `4` and `5` here per the pull request just merged will use NVM to always test the latest 4.x.x (currently 4.2.4) and 5.x.x (currently 5.3.0) version/branches respectively so until NodeJS v6 is out things should be right here for a while :smirk:

ai:
Seems like we can replace `stable` to `node`. `node` is a latest node release. What do you think about it?

ntwb:
Another note, the previous Travis build [#1095](https://travis-ci.org/postcss/autoprefixer/jobs/98911334) using `stable` was testing NodeJS v5.2.0 rather than the current v5.3.0 that is now being tested

ntwb:
On switching to `node` this also cause issues because Travis CI also use this an alias, Travis CI's alias here will be used before NVM's alias and as such using `node` will also test against NodeJS v5.2.0
, hence Travis need to update there docs surrounding all this.

You can see the `node` alias in action before and after the switch for `node-browserify` that I just submitted a pull request for also:
• https://travis-ci.org/substack/node-browserify/builds/98111654
• https://travis-ci.org/substack/node-browserify/builds/99246758

ai:
Hm, seems like we have no good solution right now :'(.

ai:
Can you also send same PR to [PostCSS](https://github.com/postcss/postcss). Autoprefixer is just a plugin and PostCSS core tests is more important.

ntwb:
Thanks @ai, thought I had, but it was https://github.com/nDmitry/grunt-postcss/pull/78, will do right now :)

ai:
`grunt-postcss`is just a small middleware between Grunt and PostCSS. PostCSS contains most of code.

ntwb:
Done, and yes, I'm pretty familiar with all the PostCSS things :+1:

PR #619, opened 2016-02-10
Add .json file-ending to all relevant require statements

### LINKED ISSUE #618: Lack of .json file-ending in prefixes.coffee necessitates jspm registry override
Hey,

I'm wondering if you would be open to a PR which adds `.json` to all require statements which interact with `caniuse-db`. @guybedford and I have been talking a bit about how to properly load autoprefixer with jspm.

It's possible to fix the issue on our end by essentially overriding how autoprefixer requires its dependencies, but a more elegant solution would be to resolve the problem at it's source.

You can see a bit more of the conversation here: https://github.com/jspm/registry/issues/749

If you're open to the idea I'll submit a PR adding the file endings ASAP.

MeoMix:
It's bad practice to implicitly load files by name without specifying
their extension. This fix helps autoprefixer work more easily with
jspm's expectations of how files should be require'd.

This resolves #618 and also respects 80-char max line length.

### DIFF
--- a/data/prefixes.coffee
+++ b/data/prefixes.coffee
@@ -45,20 +45,20 @@ add = (names..., data) ->
 module.exports = result
 
 # Border Radius
-feature require('caniuse-db/features-json/border-radius'), (browsers) ->
+feature require('caniuse-db/features-json/border-radius.json'), (browsers) ->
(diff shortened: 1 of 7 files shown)


ai:
Awesome, thank :)

MeoMix:
And thank you for the quick acceptance!! :) :+1:

PR #664, opened 2016-05-25
Repair unit conversion.

### LINKED ISSUE #663: Angle unit `rad` can not be translated into right `deg` value.
360deg should be equal to 2π rad

gucong3000:
fix #663

### DIFF
--- a/lib/hacks/gradient.coffee
+++ b/lib/hacks/gradient.coffee
@@ -72,7 +72,7 @@ class Gradient extends Value
     if /-?\d+(.\d+)?grad/.test(nodes[0].value)
       nodes[0].value = @normalizeUnit(nodes[0].value, 400)
     else if /-?\d+(.\d+)?rad/.test(nodes[0].value)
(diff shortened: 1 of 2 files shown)


lib/hacks/gradient.coffee
```diff
@@ -72,7 +72,7 @@ class Gradient extends Value
     if /-?\d+(.\d+)?grad/.test(nodes[0].value)
       nodes[0].value = @normalizeUnit(nodes[0].value, 400)
     else if /-?\d+(.\d+)?rad/.test(nodes[0].value)
-      nodes[0].value = @normalizeUnit(nodes[0].value, 2)
+      nodes[0].value = @normalizeUnit(nodes[0].value, 2 * Math.PI)
```
gucong3000: One full circle is 2π radians. **NOT** 2 radians.
https://developer.mozilla.org/en-US/docs/Web/CSS/angle
https://en.wikipedia.org/wiki/Radian

PR #725, opened 2016-10-02
Fix link to autoprefixer-rails

igoradamenko:
Links can't have same name and different URLs if you use reference text in Markdown. So because of this link to Ruby on Rails version had hash `#compass` in URL. Now it fixed.


ai:
Thanks :)

CoffeeMilkCat:
THX :+1:

PR #749, opened 2016-11-24
Hack for unitless 0 basis in IE10/IE11 shorthand flex

rzhw:
**Demo:** http://codepen.io/anon/pen/rWwaOZ
**Expected result (Chrome 55):** [Screenshot](https://cloud.githubusercontent.com/assets/79461/20585007/db4f10e8-b24b-11e6-9f89-ba120c2d28b4.png)
**Expected result (IE 11):** [Screenshot](https://cloud.githubusercontent.com/assets/79461/20585029/00e2101c-b24c-11e6-924e-15edc7ff5b6f.png)

**Overview:**

This is a workaround for [this IE10/11 bug](https://github.com/philipwalton/flexbugs#4-flex-shorthand-declarations-with-unitless-flex-basis-values-are-ignored).

Only applies when there are three components in the flex property.

This is because when there's one or two components only, it's ambiguous whether whether the first component being a unitless 0 is intended to be flex-grow or flex-basis. Browsers resolve this by treating such cases as a flex-grow of 0.

Note with IE11 that even though it understands the unprefixed flex property, we don't apply the hack there. Instead we let it treat the unprefixed version as invalid, and fall back onto -ms-flex.

Finally, although some CSS minifiers may revert the 0px to 0, we don't use 0% instead, even though many of these won't touch 0%. See [this Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=495306) for why this is the case.

**References:**

- https://www.w3.org/TR/2012/WD-css3-flexbox-20120322/#flexibility
- https://www.w3.org/TR/css-flexbox-1/#flex-property
- https://www.w3.org/TR/2016/CR-css-values-3-20160929/#component-combinators

Also adds Google Inc. to the list of copyright authors.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -48,6 +48,7 @@ Erik Sundahl <esundahl@gmail.com>
 Eugene Datsky <eugene@datsky.ru>
 Evilebot Tnawi <sheo13666q@gmail.com>
 Forrest York <https://github.com/badisa>
+Google Inc.
 Gregory Eremin <magnolia_fan@me.com>
 GU Yiling <justice360@gmail.com>
 Hallvord R. M. Steen <hallvord@hallvord.com>
--- a/lib/hacks/flex.coffee
+++ b/lib/hacks/flex.coffee
@@ -23,11 +23,17 @@ class Flex extends Declaration
     'flex'
 
   # Spec 2009 supports only first argument
+  # Spec 2012 disallows unitless basis
   set: (decl, prefix) ->
     spec = flexSpec(prefix)[0]
     if spec == 2009
       decl.value = list.space(decl.value)[0]
       decl.value = Flex.oldValues[decl.value] || decl.value
+      super(decl, prefix)
+    else if spec == 2012
+      components = list.space(decl.value)
+      if components.length == 3 and components[2] == '0'
+        decl.value = components[..1].concat('0px').join(' ')
       super(decl, prefix)
     else
       super
--- a/test/cases/flexbox.out.css
+++ b/test/cases/flexbox.out.css
@@ -212,6 +212,6 @@ a {
     -webkit-box-flex: -webkit-calc(1em + 1px);
     -webkit-flex: -webkit-calc(1em + 1px) 0 0;
        -moz-box-flex: calc(1em + 1px);
-        -ms-flex: calc(1em + 1px) 0 0;
+        -ms-flex: calc(1em + 1px) 0 0px;
             flex: calc(1em + 1px) 0 0;
 }


ai:
Good idea.

But why you add Google to `LICENSE`?

ai:
I worry because Autoprefixer have many contributors. And them are mentioned only in GitHub and ChangeLog, but not in `LICENSE`. So we need some policy for mentioned in `LICENSE`.

I never read aby good research about it. My non-professional way is to mention maintainers and developers who wrote a big part of project.

But because my thoughts is not based on good research, I am open for your opinion.

ai:
I made some research: [Rails](https://github.com/rails/rails/blob/master/MIT-LICENSE), [React](https://github.com/facebook/react/blob/master/LICENSE) and [Sass](https://github.com/sass/libsass/blob/master/LICENSE) license files mention only one person/organization.

Wikipedia [says](https://en.wikipedia.org/wiki/MIT_License), that name in `LICENSE` is not a authors, but a “copyright holders” (I understand it as person who you need to contact about this code).

rzhw:
I'll get back to you with an answer as soon as I can. Thanks!

rzhw:
One approach I've seen is to use an AUTHORS file to list copyright authors, separate from contributors:

- [google/benchmark](https://github.com/google/benchmark) has an [AUTHORS](https://github.com/google/benchmark/blob/master/AUTHORS) file and a [CONTRIBUTORS](https://github.com/google/benchmark/blob/master/CONTRIBUTORS) file for listing all contributors
- [openthread/openthread](https://github.com/openthread/openthread) has an [AUTHORS](https://github.com/openthread/openthread/blob/master/AUTHORS) which references the GitHub contributors page
- mumble-voip/mumble#2097 added an AUTHORS file and uses [a script](https://github.com/mumble-voip/mumble/blob/master/scripts/generate-AUTHORS.py) to generate it

You could try this as a starting point:

    git log --format='%aN <%cE>' | sort -u | sed -e 's/<\(.\+\?\)@users.noreply.github.com>/<https:\/\/github.com\/\1>/g' | sed -e 's/^.\+richardwa@google.com.\+$/Google Inc./' | sort

ai:
@rzhw sure, we could add `AUTHORS` with a link to GitHub? (But do we need them if main access to repo is by GitHub? :D )

Send a separated `AUTHOR` PR and remove LICENSE changes from this PR.

ai:
@rzhw ping =^_^=

rzhw:
@ai On it, will be sending a PR today 👍

ai:
OK, not please update this PR

rzhw:
Done :)

PR #777, opened 2017-01-30
Add yargs for mocha grep

jrowlingson:
Adds the [yargs](https://www.npmjs.com/package/yargs) dev dependency to allow for test filtering.

#### Example:
`gulp test --grep 'transform'`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,7 @@
 var gulp = require('gulp');
 var path = require('path');
 var fs   = require('fs-extra');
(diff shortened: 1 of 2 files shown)


ai:
Why not just use CLI `mocha [ARGS]`?

I have good alias `n` to run binaries from `node_modules/.bin` https://github.com/ai/environment/blob/master/zshrc#L52-L61

jrowlingson:
Awesome - the alias is preferable.

Issue #776, opened 2017-01-25
Regression: -webkit-box-orient rules removed


AgentME:
While upgrading my dependencies on a project with some pretty old css (and which includes [some webkit-specific hacks](https://css-tricks.com/line-clampin/)), I noticed an issue. Autoprefixer version 6.7.0 now entirely removes "-webkit-box-orient" rules. Version 6.6.1 did not have this issue.

```css
.foo {
  display: -webkit-box;
  -webkit-box-orient: vertical;
  -webkit-box-align: stretch;
}
```
->
```css
.foo {
  display: -webkit-box;
  -webkit-box-align: stretch;
}
```

Here's the code I've tested with:

```js
require('postcss')([require('postcss-cssnext')({browsers: 'chrome >= 42, safari >= 8'})]).process('.foo { display: -webkit-box; -webkit-box-orient: vertical; -webkit-box-align: stretch; }').css
```

ai:
Is it a really issue if you ask Autoprefixer to remove old prefix? ;)

IharKrasnik:
@ai How to preserve -webkit-box-orient?

ai:
Add any flexbox 2009 browser to `browserslist` in your `package.json`.

IharKrasnik:
Thank you

AgentME:
@ai This still happens even if I add `remove: false` to the options.

ai:
Are you sure that it does from Autoprefixer? Maybe some other tool (like minification in webpack’s `css-loader`) change file too?

yidea:
tested out with css-loader remain the same version, Autoprefixer is the only factor remains that causing this

AgentME:
Oh, I was passing remove:false to postcss-cssnext. Passing it on to autoprefixer results in -webkit-box-orient not being removed.

```js
var postcss = require('postcss');

var css = postcss([
  require('postcss-cssnext')({
    browsers: 'chrome >= 42, safari >= 8',
    features: {
      autoprefixer: {remove: true}
    }
  })
]).process(`
.foo {
  display: -webkit-box;
  -webkit-box-orient: vertical;
  -webkit-box-align: stretch;
}`).css;

console.log(css);
```

[...]

PR #779, opened 2017-02-01
Remove flex-flow hack

jrowlingson:
Fixes https://github.com/postcss/autoprefixer/issues/776

### DIFF
--- a/lib/processor.coffee
+++ b/lib/processor.coffee
@@ -130,8 +130,6 @@ class Processor
         notHack = @prefixes.group(decl).down (other) =>
           @prefixes.normalize(other.prop) == unprefixed
 
-        notHack = true if unprefixed == 'flex-flow'
-
         if notHack and not @withHackValue(decl)
           @reduceSpaces(decl) if decl.raw('before').indexOf("\n") > -1
           rule.removeChild(i)
--- a/test/cases/flexbox.css
+++ b/test/cases/flexbox.css
@@ -15,7 +15,13 @@ a {
 
 .a {
     display: flex;
-    flex-direction: row;
+    -webkit-box-orient: horizontal;
+    -webkit-box-direction: normal;
+    -webkit-flex-direction: row;
+       -moz-box-orient: horizontal;
+       -moz-box-direction: normal;
+        -ms-flex-direction: row;
+            flex-direction: row;
     justify-content: flex-start;
     align-items: flex-start;
     flex-wrap: nowrap;
@@ -25,7 +31,13 @@ a {
 }
 .b {
     display: flex;
-    flex-direction: row-reverse;
+    -webkit-box-orient: horizontal;
+    -webkit-box-direction: reverse;
+    -webkit-flex-direction: row-reverse;
+       -moz-box-orient: horizontal;
+       -moz-box-direction: reverse;
+        -ms-flex-direction: row-reverse;
+            flex-direction: row-reverse;
     justify-content: flex-end;
     align-items: flex-end;
     flex-wrap: wrap;
@@ -35,7 +47,13 @@ a {
 }
 .c {
     display: flex;
-    flex-direction: column;
+    -webkit-box-orient: vertical;
+    -webkit-box-direction: normal;
+    -webkit-flex-direction: column;
+       -moz-box-orient: vertical;
+       -moz-box-direction: normal;
+        -ms-flex-direction: column;
+            flex-direction: column;
     justify-content: center;
     align-items: center;
     flex-wrap: reverse-wrap;
@@ -45,7 +63,13 @@ a {
 }
 .e {
     display: flex;
-    flex-direction: column-reverse;
+    -webkit-box-orient: vertical;
+    -webkit-box-direction: reverse;
+    -webkit-flex-direction: column-reverse;
+       -moz-box-orient: vertical;
+       -moz-box-direction: reverse;
+        -ms-flex-direction: column-reverse;
+            flex-direction: column-reverse;
     justify-content: space-between;
     align-items: baseline;
     align-content: space-between;


ai:
Nice try :D. You broke the tests ;).

You need to fix it without test changes.

ai:
I added this hack to fix prefixes removing. We can’t just remove hack, because in this case Autoprefixer will not be able to remove prefixes.

jrowlingson:
Ah thanks for taking a look :)


Correct me if I am wrong, but, if I am thinking about the fix correctly we will specifically want to retain the `-webkit-box-orient` declaration when present - so I would still need to change the tests for this case.

ai:
Why we should keep `-webkit-box-orient`?

Let’s me explain Flexbox tests:

1. Test that prefixes added correctly: `flexbox.css` → AP → `flexbox.out.css`
2. Test that we remove all prefixes: `flexbox.out.css` → AP(`browsers: []`) → `flexbox.css`
3. Test that we do not double prefixes: `flexbox.css` → AP → AP → `flexbox.out.css`
4. Test that we can disable flexbox: `flexbox.out.css` → AP(`flexbox: false`) → `flexbox.css`

I added `notHack = true if unprefixed == 'flex-flow'` to fix test number 2.

Autoprefixer try to find a groups:

```css
.foo {
  -webkit-box-shadow: x;
  box-shadow: x
}
```

If he didn’t find a unprefixed (`box-shadow`) version in group, it decide that it is the “hack”, and should not be removed.

Unfortunately, in this case we need 2-3 different prefixed properties:

```css
.foo {
  -webkit-box-orient: horizontal;
    -webkit-box-direction: reverse;
            flex-direction: row-reverse;
}
```

Because `box-direction` is not equal to `flex-direction, Autoprefixer decide, that `-webkit-box-direction` is a hack (prefix without unprefixed version near by). So removing test was broken.

This is why I remove hack detection logic for this case (honestly, because I gave up, logic became too complicated here, Autoprefixer should not be perfect in any case).

jrowlingson:
Thanks for the explanation; the behavior the tests are asserting is clear to me now. I see that the tests should not need to change here.

> Why we should keep -webkit-box-orient?

The original issue is that there are valid cases where the `-webkit-box-orient` is used as a hack so removing it causes the layout to break. For example, when used as a [line clamp](https://css-tricks.com/line-clampin/#article-header-id-0): 

```css
.line-clamp {
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;  
}
```

But I can see the hack detection logic needed here is non-trivial. I might take another pass at detecting this case if you think its worth it.

jamesbrobb:
I have the same use case for  `-webkit-box-orient: vertical;, so am also now experiencing issues with broken layouts, due to the removal.

ai:
@jamesbrobb could you show you more of your CSS?

jamesbrobb:
Apologises, i thought i was running the latest version of chrome, in which the following css still worked:
```
{
    display: -webkit-box;
    -webkit-line-clamp: 2;
    -webkit-box-orient: horizontal;
}
```
But have just updated to  `56.0.2924.87` and see that these styles are no longer valid.

jamesbrobb:
Sorry, my mistake again, i was testing and hadn't noticed i had `horizontal` instead of `vertical`.

This still works in the latest version of chrome and as i'm using 'last 2 versions' shouldn't it still be supported and therefore not removed?

.clamp-text {
    display: -webkit-box;
    -webkit-line-clamp: 2;
    -webkit-box-orient: vertical;
}

ai:
Nope, `-webkit-` prefixes is not required for `last 2 version` http://caniuse.com/#search=flexbox

ai:
Also I highly recommend to refector this code and replace `display: -webkit-box` to `display: flex`.

You have a Autoprefixer for prefixes, manually written prefixes is a bad practices. Prefixes for only one browsers is not really moral thing ;). Prefixes without unprefixed version is not future-proof — Chrome will disable `-webkit-box` in some future and you code will be broken.

romulof:
`-webkit-line-clamp` only works with `-webkit-box`

domschmidt:
We're using line clambing if the browser supports it, e.g. as follows:

```
  @supports (-webkit-line-clamp: 1) {
    display: -webkit-box;
    -webkit-line-clamp: 1;
    -webkit-box-orient: vertical;
  }
```

As a result, -webkit-box-orient: vertical; is removed at all

ai:
Disable Autoprefixer here by control comments or use `remove: false` option.

domschmidt:
Works fine, thank you

RVNranjan:
@ai im using gulp/grunt inside node_modules\autoprefixer\node_modules\postcss\lib , i have postcss.js but not postcssrc.js

ai:
@RVNranjan show your config for gulp

RVNranjan:
@ai 

'use strict';

var gulp   = require('gulp'),
    help   = require('gulp-help')(gulp),
    config = require('./lib')(gulp, { help: true });

// ===================
// CONFIG
// ===================

config({

    paths: {
        source: ['./lib/**/*.js', './tasks/**/*.js'],
        tests : ['./test/**/*.js']
    },

    // tasks
    // ------------------------

    jshint  : { lint : '<%=paths.source%>' },
    mocha   : { specs: '<%=paths.tests%>'  }

});

gulp.task('test', ['jshint', 'mocha']);

ai:
@RVNranjan do you already have Autoprefixer in your project? Or you all me how to add it?

RVNranjan:
I already have that

RVNranjan:
@ai I already have Autoprefixer in my project, just i want to know how to do Autoprefixer return: false.

ai:
@RVNranjan OK, where did you enable the Autoprefixer?

In Grunt config? Or you use it as text editor plugin?

RVNranjan:
@ai autoprefixer is enabled in Grunt config

ai:
@RVNranjan awesome! :) show a config and I will tell where exactly you need to change it :)

PR #785, opened 2017-02-08
Fixed a typo in the code of readme

dariusrosendahl:
Exclamation mark was missing with autoprefixer: off. Without it, it doesn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,7 +494,7 @@ a {
 }
 
 b {
-    /* autoprefixer: off */
+    /*! autoprefixer: off */
     transition: 1s; /* it will not be prefixed */
 }
 ```


ai:
It works without `!`

https://github.com/postcss/autoprefixer/blob/master/test/cases/disabled.css#L8

You need `!` because of Sass — on compiling it removes all comments. `!` in comment will prevent Sass from removing comment.

But `!` will not work for non-Sass environments. SO we couldn’t recommend only `!`.

So, could you make other PR, with note about Sass and `!`?

dariusrosendahl:
Ah ok, that explains it. I'll make another PR

PR #793, opened 2017-02-21
Clip Path should derive from css-clip-path feature

cvle:
I think clip-path should derive from `caniuse-db/features-json/css-clip-path.json` instead of `caniuse-db/features-json/css-masks.json`.
This was a real quick patch. Please check for correctness.


ai:
Looks great. Sorry, I will deal with it only tomorrow. Is it OK?

cvle:
Yes don't worry. Take your time. You can probably expect more PRs coming :-)

PR #808, opened 2017-03-29
Fix postcss-cli usage

RyanZim:
The old example won't work in postcss-cli v3, since `--use` takes an array.


RyanZim:
@ai ping?

ai:
Oops, I missed first comment. It is good that you pinged me.

PR #828, opened 2017-04-28
Adding hyphenate prefixes

KurtPreston:


KurtPreston:
Huh, not sure why that build failed.  I figured it was related to the `caniuse` version bump, but `gulp test` works fine locally?

ai:
Can you send PR to `v7` branch? Everything was rewritten since it.

Second question: how do you cover the case, when you Firefox supports only few properties from spec? You add prefixes for all properties?

ai:
Can you show input and output CSS for this changes? I checked Can I Use PR and didn't find code for prefixes there.

Issue #803, opened 2017-03-15
[css-grid]min-content/max-content prefix in Safari


yisibl:
input:

```css
.foo {
     grid-template-columns: minmax(100px, min-content);
}
```

output:

```css
.foo {
     -ms-grid-columns: minmax(100px, min-content);
         grid-template-columns: minmax(100px, min-content);
}
```

These properties just remove the prefix in https://trac.webkit.org/changeset/213831 , but I'm not sure which version of Safari.

@mrego Can you help me?

http://caniuse.com/#feat=css-grid

mrego:
Apple doesn't publicly talk about their Safari releases, so we don't know.
I don't think it would be part of 10.1 as the beta has been around for a while, so probably Safari 11 or 10.2 if that ever exists.
Sorry for not being really helpful, you'd need to wait and check when it's or not included.

ai:
@yisibl what is expected output?

yisibl:
@mrego Well, it looks like caniuse's data is inaccurate.

@ai Maybe we don't need a prefix in `minmax(100px, -webkit-min-content)`,We have to wait for Safari to officially support Grid.

ai:
@yisibl ouh, I will add grid properties to `min-content` https://github.com/postcss/autoprefixer/blob/master/data/prefixes.coffee#L292

Do you want to make PR to save your name in history (again)? :)

mrego:
One special thing here, `fit-content` cannot be prefixed inside grid properties, or it won't work. I know it's a weird situation but that's how things are right now.
So valid declarations would be:
```grid-template-columns: -webkit-min-content -webkit-max-content fit-content(500px);```
However the following one would be invalid:
```grid-template-columns: -webkit-min-content -webkit-max-content -webkit-fit-content(500px);```

Note that `fit-content` in the grid properties is not a keyword like `min|max-content` but a kind of "function". That's the reason why it was implemented as something new and unprefixed since the beginning.

yisibl:
@mrego OK, I see, `fit-content()` no prefix.

@ai  Yeah, willing to serve you 👻 .

yisibl:
Safari 10.1 shipped: https://developer.apple.com/library/prerelease/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_1.html#//apple_ref/doc/uid/TP40014305-CH12-DontLinkElementID_55

yisibl:
[...]

PR #840, opened 2017-05-15
Fix: Do not add non-webkit prefixes for CSS grid

yisibl:
In CSS grid, only Safari needs to add the -webkit- prefix to min-content/max-content values.

Fixed this commit: https://github.com/postcss/autoprefixer/commit/99def371c9511212dfc9c1eaa573abfee813b97d?diff=split#diff-edefb210647f5ec99168718dbbf80320R36

and this issue: https://github.com/postcss/autoprefixer/issues/803

@ai


ai:
Thanks! Great PR.

PR #888, opened 2017-09-02

regularlabs:
Changes if/else structure (flips it) so early returns can be used and else can be removed.
Result: more readable, easier to follow


ai:
Do you have Twitter? I what to tweet that code cleaning is a good idea for open source contribution

regularlabs:
You can follow me on @RegularLabs :)

PR #889, opened 2017-09-02

regularlabs:
This one has a bit more refactoring going on, so you will need a good test run n this. Or drop it, if you don't think it is worth it :)

It would clean stuff up a bit to extract a large part of the code in the for loop to a separate getItem method.
That way you can do early returns more easily and only need one result.push(item);

Something like:
``` 
                item = this.getItem(propPrefix, decl);
                result.push(item);
```

### DIFF
--- a/lib/value.js
+++ b/lib/value.js
@@ -93,9 +93,9 @@ class Value extends Prefixer {
     value(decl) {
         if (decl.raws.value && decl.raws.value.value === decl.value) {
             return decl.raws.value.raw;
+        } else {
+            return decl.value;
         }
-
-        return decl.value;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lib/value.js
```diff
@@ -82,9 +93,9 @@ class Value extends Prefixer {
     value(decl) {
         if (decl.raws.value && decl.raws.value.value === decl.value) {
             return decl.raws.value.raw;
-        } else {
-            return decl.value;
         }
```
ai: In this example `if return - else return` is more readable in my opinion

ai:
Revert that commented part and I will accept PR :)

regularlabs:
Done

PR #891, opened 2017-09-02

regularlabs:
Removes redundant elses after early returns.
Result: less indentation, more readable, easier to follow

### DIFF
--- a/lib/transition.js
+++ b/lib/transition.js
@@ -117,29 +117,31 @@ class Transition {
      * Show transition-property warning
      */
     checkForWarning(result, decl) {
-        if (decl.prop === 'transition-property') {
-            decl.parent.each((i) => {
-                if (i.type !== 'decl') {
-                    return undefined;
-                }
-                if (i.prop.indexOf('transition-') !== 0) {
-                    return undefined;
-                }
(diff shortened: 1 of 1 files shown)


lib/transition.js
```diff
@@ -173,9 +175,10 @@ class Transition {
 
         if (double || smaller) {
             decl.remove();
-        } else {
-            decl.value = value;
+            return;
         }
+
+        decl.value = value;
```
ai: Not sure about this 2 lines case. `if-else` tree show that it is two ways.

regularlabs: I like to think of a 'golden path' in a method.
You want a method to do one thing.
All the ifs 'should' check if there are cases that are other than the ideal situation (like exceptions) in which case you want it to do something else (like pass it to a different method or do an early return or such).
If all the 'negative' ifs can be skipped, then you get to the statement you actually want to do. That statement should pretty much also be doing what the method name describes.

Of course, this doesn't/can't apply in all cases, but with that mindset it is much easier to write small and readable methods that have no complicated if/else structures with 16 levels of indentation.

That's just to explain my thinking behind it. If you prefer a different approach, that is of course completely up to you! :)

PR #894, opened 2017-09-02
[Refactor] Changes if/else structure to switch statement

regularlabs:
I think this is more more readable and easier to follow.
I you disagree, dump it :)


ai:
I big fan of `if` :)

PR #916, opened 2017-10-05
fix little syntax

RayJune:


ai:
Thanks 😊 but it is not a typo, `npx` is a new tool from npm 5 to run commands from npm dependencies without installing them globally.

PR #928, opened 2017-10-20
Add code coverage, tests of initializing and remove not used code

Semigradsky:
Code coverage
===

Before:
--
> 97.12% Statements 2093/2155
> 92.69% Branches 1192/1286
> 97.45% Functions 382/392
> 97.23% Lines 2069/2128

After:
--
> 97.67% Statements 2099/2149
> 93.23% Branches 1199/1286
> 97.95% Functions 382/390
> 97.79% Lines 2075/2122

### DIFF
--- a/lib/autoprefixer.js
+++ b/lib/autoprefixer.js
@@ -92,6 +92,8 @@ module.exports = postcss.plugin('autoprefixer', (...reqs) => {
 
     plugin.options = options;
 
+    plugin.browsers = reqs;
+
     plugin.info = function (opts) {
         opts = opts || {};
         opts.from = opts.from || process.cwd();
--- a/test/autoprefixer.test.js
+++ b/test/autoprefixer.test.js
@@ -145,9 +145,31 @@ it('throws on wrong options', () => {
     }).toThrowError(/browsers/);
 });
 
-it('sets options', () => {
-    const opts = { browsers: ['chrome 25', 'opera 12'], cascade: false };
-    expect(autoprefixer(opts).options).toEqual(opts);
+const options = {
+    cascade: false,
+    grid: false
+};
+
+const browsers = ['chrome 25', 'opera 12'];
+
+it('sets options via options object', () => {
+    const allOptions = Object.assign(options, { browsers });
+
+    const instance = autoprefixer(allOptions);
+    expect(instance.options).toEqual(allOptions);
+    expect(instance.browsers).toEqual(browsers);
+});
+
+it('sets options via array of browsers as first argument and object', () => {
+    const instance = autoprefixer(browsers, options);
+    expect(instance.options).toEqual(options);
+    expect(instance.browsers).toEqual(browsers);
+});
+
+it('sets options via browsers as arguments and options object', () => {
+    const instance = autoprefixer(...browsers, options);
+    expect(instance.options).toEqual(options);
+    expect(instance.browsers).toEqual(browsers);
 });
 
 it('has default browsers', () => {
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 node_modules/
 
 build/
+coverage/
--- a/.npmignore
+++ b/.npmignore
@@ -4,6 +4,7 @@
 node_modules
 yarn.lock
 build
+coverage
 autoprefixer.js
 
 test
--- a/lib/browsers.js
+++ b/lib/browsers.js
@@ -56,21 +56,6 @@ class Browsers {
     }
 
     /**
-     * Select major browsers versions by criteria
-     */
-    browsers(criteria) {
-        let selected = [];
-        for (const browser in this.data) {
-            const data = this.data[browser];
-            const versions = criteria(data).map(
-                version => `${browser} ${version}`
-            );
-            selected = selected.concat(versions);
-        }
-        return selected;
-    }
-
-    /**
      * Return prefix for selected browser
      */
     prefix(browser) {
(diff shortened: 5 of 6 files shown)


lib/prefixes.js
```diff
@@ -51,7 +51,6 @@ Value.hack(require('./hacks/intrinsic'));
 Value.hack(require('./hacks/pixelated'));
 Value.hack(require('./hacks/image-set'));
 Value.hack(require('./hacks/cross-fade'));
-Value.hack(require('./hacks/flex-values'));
```
ai: Are you sure about this changes? Looks dangerous.

lib/autoprefixer.js
```diff
@@ -28,40 +28,52 @@ function timeCapsule(result, prefixes) {
     );
 }
 
-module.exports = postcss.plugin('autoprefixer', (...reqs) => {
+function getOptions(...args) {
```
ai: Why not just `getOptions(args)`? Faster and simpler ;).

lib/autoprefixer.js
```diff
@@ -28,40 +28,52 @@ function timeCapsule(result, prefixes) {
...
-    } else if (reqs.length === 0 || reqs.length === 1 && !reqs[0]) {
-        reqs = undefined;
-    } else if (reqs.length <= 2 && (reqs[0] instanceof Array || !reqs[0])) {
-        options = reqs[1];
-        reqs = reqs[0];
-    } else if (typeof reqs[reqs.length - 1] === 'object') {
-        options = reqs.pop();
+    let reqs;
+
+    // autoprefixer([...,] options)
```
ai: I prefer to not write this type of comments. If you need comment, you should make code clearly.

lib/prefixes.js
ai: Yeap, it is dangerous. We use it for `transition: flex-grow`.

Semigradsky: Oh I see... I have returned file.
Would be good to add test for this case.

lib/autoprefixer.js
Semigradsky: Exactly!

lib/autoprefixer.js
Semigradsky: I have rewrote a little.

ai:
I don’t like many changes in the code. We had bad story in Browserslist when coverage fixes broke release.

What do you think of not changing origin code and only add new tests?

Semigradsky:
Ok, I will change PR. I will ping when it be done.

Semigradsky:
@ai ping! :point_left:

test/autoprefixer.test.js
```diff
@@ -145,9 +145,33 @@ it('throws on wrong options', () => {
     }).toThrowError(/browsers/);
 });
 
-it('sets options', () => {
-    const opts = { browsers: ['chrome 25', 'opera 12'], cascade: false };
-    expect(autoprefixer(opts).options).toEqual(opts);
+describe('sets options', () => {
```
ai: You don’t need `description` here. Also it double the test name

test/autoprefixer.test.js
```diff
@@ -145,9 +145,33 @@ it('throws on wrong options', () => {
...
+    const opts = {
+        browsers: ['chrome 25', 'opera 12'],
+        cascade: false,
+        grid: false
+    };
+
+    it('sets options (1)', () => {
```
ai: `1, 2, 3` is bad tests names

lib/autoprefixer.js
```diff
@@ -58,6 +58,8 @@ module.exports = postcss.plugin('autoprefixer', (...reqs) => {
 
     if (options.browsers) {
         reqs = options.browsers;
+    } else {
```
ai: Why we need this change?

Semigradsky: If we want have `browsers` in `options`.

ai: But we don’t want to change API ;)

Semigradsky: I'm not sure how check setting `browsers` without getting it from `options` after initialize.

ai: Why you even need it? You improve tests, yeap? Not a API?

ai: Ouh, you mean that you need to get target browsers for tests? Changing original `options`  object is a bad way to do it.

You can use `info()` or check output prefixes.

Semigradsky: So weird. We have specified the browsers but can't get him by `instance.options`

ai: Or you can add special `plugin.browsers` with browsers after `browserslist`

test/autoprefixer.test.js
```diff
@@ -145,9 +145,33 @@ it('throws on wrong options', () => {
     }).toThrowError(/browsers/);
 });
 
-it('sets options', () => {
-    const opts = { browsers: ['chrome 25', 'opera 12'], cascade: false };
-    expect(autoprefixer(opts).options).toEqual(opts);
+describe('sets options via', () => {
```
ai: We don’t use this type of grouping in other tests. It will be better to remove `describe` in terms of tests consistency

Semigradsky:
Added `plugin.browsers`

lib/autoprefixer.js
```diff
@@ -92,6 +92,8 @@ module.exports = postcss.plugin('autoprefixer', (...reqs) => {
 
     plugin.options = options;
 
+    plugin.browsers = reqs;
```
ai: It is be better to put not a `reqs` here, but browsers after Browserslist ;)

Issue #931, opened 2017-10-26
info() not seeing browserlist configuration


pdokas:
This is similar to #838.

When using 7.1.6, the `info()` method on the plugin returns the default list of browsers no matter what your configuration is.

Example usage:

**.browserslistrc** (or `browserslist`, or `package.json`)
```
safari >= 6
```

**gulpfile.js**
```js
var autoprefixer = require('autoprefixer');

var ap = autoprefixer({
	cascade: false,
	add: true,
	remove: false,
	supports: true,
	flexbox: true,
	grid: false
});

console.log(ap.info());
```

Results in this output:

```
Browsers:
  Chrome for Android: 61
  Firefox for Android: 56
  And_qq: 1.2
  UC for Android: 11.4
  Android: 56, 4.4.3-4.4.4
  Baidu: 7.12
  Bb: 10, 7
  Chrome: 62, 61, 60, 49
  Edge: 16, 15
  Firefox: 56, 55, 52
  IE: 11, 10
  IE Mobile: 11, 10
  iOS: 11, 10.3, 10.0-10.2
  Opera Mini: all
  Opera Mobile: 37, 12.1
  Opera: 48, 47
  Safari: 11, 10.1
  Samsung: 5, 4
```

Where the correct output would look like:

```
Browsers:
  Safari: 11, 10.1, 10, 9.1, 9, 8, 7.1, 7, 6.1, 6
```

Looking into the code, I believe the issue is that the [`plugin.info`](https://github.com/postcss/autoprefixer/blob/7.1.6/lib/autoprefixer.js#L95-L97) method’s `opts` var is `undefined` since nothing is passed in ([as per the docs](https://github.com/postcss/autoprefixer#debug)).

I believe the fix is to change [line 96](https://github.com/postcss/autoprefixer/blob/7.1.6/lib/autoprefixer.js#L96) to be:

```js
return require('./info')(loadPrefixes({ from: path.resolve('.') }));
```

I’d be happy to make a PR to this effect, but I wanted to open a discussion first!

ai:
You must provide `from` option

```js
console.log(ap.info({ from: process.cwd() }));
```

But I understand that it is not clear. There are 2 options:

1. Add default value (`process.cwd`) to `from`. But if users will have different configs (like `src/legacy/.browserslistrc`), they will not know how to load it properly.
2. Show warning on missed `from` option.

What do you think?

pdokas:
[...]

PR #932, opened 2017-10-27
Provide a default file path for info() output

pdokas:
Fixes https://github.com/postcss/autoprefixer/issues/931.

This provides a default file path when invoking the `info()` method on the plugin. This allows the plugin to, by default, retrieve the local config from the file system. If you pass a hash to `info`, then that will be used without modification.

### DIFF
--- a/lib/autoprefixer.js
+++ b/lib/autoprefixer.js
@@ -93,6 +93,9 @@ module.exports = postcss.plugin('autoprefixer', (...reqs) => {
     plugin.options = options;
 
     plugin.info = function (opts) {
+        opts = opts || {};
+        opts.from = opts.from || process.cwd();
+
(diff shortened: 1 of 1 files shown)


louisbuchbinder:
👍

ai:
Thanks. But also we need to cover `info({ })` case too :)

pdokas:
@ai Would it be better if I made sure `opts.from` itself were always specified?

ai:
@pdokas yeap.

pdokas:
I’ve rebased that in!

ai:
I will release this PR together with Grid props optimization (somebody should fix it in the week).

ai:
Released in 7.2

Issue #883, opened 2017-08-28
support grid-area, grid-row-end, and grid-column-end with span syntax


hellochar:
autoprefixer seems to completely ignore `grid-area`, `grid-row-end`, and `grid-column-end`, even though the `span` syntax should be translatable:

```css
grid-area: 5 / 1 / span 1 / span 5;

/* could turn into */

-ms-grid-row: 5;
-ms-grid-column: 1;
-ms-grid-row-span: 1;
-ms-grid-column-span: 5;
```

ai:
Sure. Do you want to make PR? I will not have on this week.

ai:
https://github.com/postcss/autoprefixer/pull/938 improved Grid support

ai:
https://github.com/postcss/autoprefixer/pull/943 improved Grid support even better

ai:
Even `grid-template-areas` support was added https://github.com/postcss/autoprefixer/pull/948

ai:
Released in 7.2

PR #938, opened 2017-11-08
Add partial grid-area support, improve grid-row and grid-column

Alex7Kom:
Another take on #883 

While the original issue suggested only translation of span syntax, this is an attempt to translate `grid-area` shorthand to `-ms-` properties as fully as possible. It is also applied to `grid-row` and `grid-column` shorthands.

~~Tests are failing:~~
~~1) 'removes unnecessary prefixes': presumably because it also should remove `-ms-` properties and currently it doesn't.~~
~~2) 'prevents doubling prefixes': it adds properties again.~~

~~Currently I'm trying to figure it out.~~

### DIFF
--- a/lib/hacks/grid-area.js
+++ b/lib/hacks/grid-area.js
@@ -1,0 +1,74 @@
+const Declaration = require('../declaration');
+const shorthand = require('./grid-shorthand');
+
+class GridArea extends Declaration {
+
+    static names = [
+        'grid-area'
+    ];
+
+    /**
+     * Check if -ms-grid-row is already present
+     */
+    checkIfAlreadyApplied(decl) {
+        for (const i in decl.parent.nodes) {
+            if (decl.parent.nodes.hasOwnProperty(i)) {
+                const element = decl.parent.nodes[i];
+                if (element.prop === '-ms-grid-row') {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    /**
+     * Translate grid-area to separate -ms- prefixed properties
+     */
+    insert(decl, prefix, prefixes) {
+        if (prefix !== '-ms-') {
+            return super.insert(decl, prefix, prefixes);
+        }
+
+        if (!this.checkIfAlreadyApplied(decl)) {
+            return undefined;
+        }
+
+        const values = shorthand.parse(decl);
+
+        const [rowStart, rowSpan] = shorthand.translate(values, 0, 2);
+        const [columnStart, columnSpan] = shorthand.translate(values, 1, 3);
+
+        if (rowStart) {
+            decl.cloneBefore({
+                prop: '-ms-grid-row',
+                value: rowStart.toString()
+            });
+        }
+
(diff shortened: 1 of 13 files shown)


ai:
Tests fall. Seems like you fix can’t clean own prefixes. In this case, you can create separated test case just for `grid-area` and do not use this tests in `removes unnecessary prefixes`.

Alex7Kom:
Done! I also have figured out 'doubling prefixes' check.
Now it fails on size limit, not sure what to do about it.

ai:
@Alex7Kom just increase limit in `package.json/size-limit` :)

lib/hacks/grid-area.js
```diff
@@ -0,0 +1,72 @@
+const parser = require('postcss-value-parser');
+
+const Declaration = require('../declaration');
+const translateShorthand = require('./grid-shorthand');
```
ai: If you use it only in the single file, it will be better to put `./grid-shorthand.js` content to `grid-area.js`.

lib/hacks/grid-area.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return true;
+    }
+
+    /**
+     * Translate grid-area to separate -ms- prefixed properties
+     */
+    insert(decl, prefix) {
+        if (prefix !== '-ms-') {
+            return;
```
ai: I think

```js
return super.insert(decl, prefix)
```

is more future-proof :).

lib/hacks/grid-area.js
```diff
@@ -0,0 +1,72 @@
...
+                values[current] = [];
+            } else if (i.type === 'word') {
+                values[current].push(i.value);
+            }
+        }
+
+        const [rowStart, rowSpan] = translateShorthand(values, 0, 2);
+        const [columnStart, columnSpan] = translateShorthand(values, 1, 3);
+
+        rowStart && decl.cloneBefore({
```
ai: `if (rowStart)` is more simple construction. It will be better to future developers.

lib/hacks/grid-area.js
```diff
@@ -0,0 +1,72 @@
...
+    ];
+
+    /**
+     * Check if -ms-grid-row is already present
+     */
+    check(decl) {
+        for (const i in decl.parent.nodes) {
+            if (decl.parent.nodes.hasOwnProperty(i)) {
+                const element = decl.parent.nodes[i];
+                if (element.prop === '-ms-grid-row') {
```
ai: Not future-proof. It is better to have this test in `insert`.

Alex7Kom:
Fixed everything noted, also added `grid-row` and `grid-column` translation.

ai:
Awesome!

ai:
Released in 7.2

Ayhanbey63500 (commented):
[Gitkap](https://www.github.com/sindresorhus/xo/tree/master/)

Green369-1 (approved):

PR #939, opened 2017-11-08
Add a Git .mailmap with my new name

mgol:
In this way my past contribution is mapped correctly.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>


ai:
Can you tell me more about this file and what does it fix?

mgol:
`.mailmap` is used by `git shortlog` (see e.g. what `git shortlog -nse` prints) & `git blame`. By default it doesn't affect `git log` or `git show`, mainly for historical reasons (see http://git.661346.n2.nabble.com/git-log-to-use-mailmap-by-default-td7570862.html). It can be made so in two ways:

1. By adding a `--use-mailmap` option: `git log --use-mailmap` or `git show COMMIT_HASH --use-mailmap`
2. By configuring Git globally to respect the `.mailmap` for logging: `git config --global log.mailmap true`. When set, this option is also respected for `git show`.

`.mailmap` affects what lots of apps outside of the Web GitHub UI present (mainly because Git respects it in many places). For example, it changes what WebStorm shows when you use its "Annotate" feature - a visual representation of `git blame` inside of the editor.

If you'd like, I can spend a moment to map other people's data; see e.g. my PR to `html-webpack-plugin`: https://github.com/jantimon/html-webpack-plugin/pull/794.

ai:
1. What does this PR will change on GitHub?
2. Did you change your email? Or your previous commits had the same email but doesn’t have a name?

mgol:
> 1\. What does this PR will change on GitHub?

GitHub doesn't take `.mailmap` into account, relying instead on its own accounts (and ignoring this issue for contributors without a GitHub account). This matters for tools outside of GitHub - the command line Git, various Git UIs like [SourceTree](https://www.sourcetreeapp.com/) or [Tower](https://www.git-tower.com/) and IDEs like [WebStorm](https://www.jetbrains.com/webstorm/).

> 2\. Did you change your email? Or your previous commits had the same email but doesn’t have a name?

I changed my name. Previously I was committing under the same email but a different name.

There's also a funny story in that `ę` can be written in two different ways that look the same in most editors and some of those editors convert it one way or the other which (without a `.mailmap` entry) makes me counted as two different people in most projects where I have multiple commits... See [this PR to Karma](https://github.com/karma-runner/karma/pull/2788) for more details about this aspect.

ai:
So what will be your benefits from this PR if GitHub will be the same and other Git tools are used mostly by me during development? :D

I am not big fan of many dot-files and we need a reason to add it.

mgol:
If you put it that way... :)

Feel free to close it, no biggie. I just prefer to have my persona unified if I'm checking some stats or doing blame when checking some piece of code but, obviously, the usage of all that will be very limited since you're working on the code, not me. So if you prefer to avoid that extra file, no problem.

ai:
I thought about it and still think to close this PR.

I could accept it if it will change something on GitHub or any other important website. But since we don’t have many contributors in Autoprefixer in this days, I think nobody will see the changes. In this case, we use Occam’s razor — less configs is better.

PR #961, opened 2017-12-30
Support for classes in IE 10 and below

AleshaOleg:
To support ES6 classes in IE 10 and below, according to Babel [docs](https://babeljs.io/docs/usage/caveats/#internet-explorer-classes-10-and-below-) projects need to use `babel-plugin-transform-es2015-classes` (https://www.npmjs.com/package/babel-plugin-transform-proto-to-assign). So, I added it on this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.0.3",
     "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-es2015-classes": "^6.24.1",
     "babel-preset-env": "^1.6.1",
     "babelify": "^8.0.0",
     "browserify": "^14.5.0",
@@ -105,6 +106,12 @@
         "transform-class-properties",
         {
           "spec": true
+        }
+      ],
+      [
+        "transform-es2015-classes",
+        {
+          "loose": true
         }
       ]
     ]
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "babel-eslint": "^8.0.3",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-es2015-classes": "^6.24.1",
+    "babel-plugin-transform-proto-to-assign": "^6.26.0",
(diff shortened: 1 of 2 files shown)


ai:
This manually way is not recommended by Babel team. We already use `babel-preset-env` which automatically choice plugging according Browserslist queries.

So, if you want to support IE 10 you just need to update target browser query (current is `last 1 version`, IE 10 is out of this query).

Why do we need to support IE 10?

AleshaOleg:
Does not `autoprefixer` need to support as many browsers as possible? This need to make `postcss-js` working in old browsers.

ai:
I think running Autoprefixer in the browser is not the best way 😅

I think only about Codepen apps. They don't need IE 10.

What is your user case?

AleshaOleg:
I'm using `postcss-js` in [css-vendor](https://github.com/cssinjs/css-vendor) to test it. I'm want to update `autoprefixer` to last version there to get all last changes for CSS properties. And tests fails of IE 9/10 (https://travis-ci.org/cssinjs/css-vendor/builds/323397048?utm_source=github_status&utm_medium=notification) because `hack` method from `autoprefixer` not extending properly from ES6 classes:( IE9/10 not supports - `setPrototypeOf` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf). `setPrototypeOf` using by Babel to extends `_proto_` of current class from another one.
Could we please make a support for this case?

ai:
I will think about it when I will be sober 😅

AleshaOleg:
Haha, ok😆 Happy new year 🎉 🎅🏻

With best regards, Oleh Aloshkin.
Email: olegaleshkin@gmail.com
Mobile phone: +380 63 282 33 81
Skype: aleshaoleg

ai:
Done fe3c73d

I will release new PostCSS version and then Autoprefixer version.

ai:
I added IE 10 support to PostCSS too https://github.com/postcss/postcss/commit/10d4e649c5f14ec3295f0130e174d6c1e9ab79d6

ai:
Released in 7.2.10

PR #971, opened 2018-01-18
docs(README): add install part

kud:


ai:
We need separating Install instructions for each environment (Gulp, Grunt, etc) from Usage section.

For example, for Gulp you need to install `gulp-postcss` and `autoprefixer` to use it.

Do you want to update pull request?

PR #976, opened 2018-01-28
Can't find browserslist config from browserslist config file or packa…

ken1987:
My english is poor，I'll give some code and comment to describe the question

## step1: [autoprefixer/lib/autoprefixer.js#L81](https://github.com/postcss/autoprefixer/blob/f37e5ba9cff5c2d4429306b3f0562f130c7b6d6c/lib/autoprefixer.js#L81)

```
const plugin = function (css, result) {
        const prefixes = loadPrefixes({
            // sometimes `css.source && css.source.input.file === undefined`
            from: css.source && css.source.input.file,  
            env: options.env
        });
        timeCapsule(result, prefixes);
        if (options.remove !== false) {
            prefixes.processor.remove(css, result);
        }
        if (options.add !== false) {
            prefixes.processor.add(css, result);
        }
    };
```

## step2: [browserslist/index.js#L189](https://github.com/ai/browserslist/blob/e3e916f663e6bac7504ddc0ceeb5ec37d7c496a8/index.js#L189)

```js
function browserslist (queries, opts) {
  if (typeof opts === 'undefined') opts = { }

  if (!opts.hasOwnProperty('path')) {  // It is false, opts={ path: undefined, env: undefined }                      
    opts.path = path.resolve ? path.resolve('.') : '.'
  }
   
  // when runned here, opts.path is still equal to undefined
}
```

## step3: [browserslist/node.js#L115](https://github.com/ai/browserslist/blob/e3e916f663e6bac7504ddc0ceeb5ec37d7c496a8/node.js#L115)

```js
loadConfig: function loadConfig (opts) {
    if (process.env.BROWSERSLIST) {
      return process.env.BROWSERSLIST
    } else if (opts.config || process.env.BROWSERSLIST_CONFIG) {
      var file = opts.config || process.env.BROWSERSLIST_CONFIG
      if (path.basename(file) === 'package.json') {
        return pickEnv(parsePackage(file), opts)
      } else {
        return pickEnv(module.exports.readConfig(file), opts)
      }
    } else if (opts.path) {    // It is undefined
      return pickEnv(module.exports.findConfig(opts.path), opts)
    } else {
      return undefined       // get it
    }
  }
```

### DIFF
--- a/lib/browsers.js
+++ b/lib/browsers.js
@@ -48,11 +48,15 @@ class Browsers {
      * Return browsers selected by requirements
      */
     parse(requirements) {
-        return browserslist(requirements, {
+        const {from, env} = this.options
+        const options = {
             stats: this.stats,
-            path: this.options.from,
-            env: this.options.env
-        });
+            env
+        }
+        if(from) {
+            options.path = from
+        }
+        return browserslist(requirements, options);
     }
 
     /**
--- a/lib/browsers.js
+++ b/lib/browsers.js
@@ -48,13 +48,13 @@ class Browsers {
      * Return browsers selected by requirements
      */
     parse(requirements) {
-        const {from, env} = this.options
+        const { from, env } = this.options;
         const options = {
             stats: this.stats,
             env
-        }
-        if(from) {
-            options.path = from
+        };
+        if (from) {
+            options.path = from;
         }
         return browserslist(requirements, options);
     }


ai:
I just came from 10 hours flight. So I will check it only tomorrow.

But looks very promising ;).

ai:
The question — maybe we should remove `opts.hasOwnProperty('path')` from Browserslist. Tomorow I will try to remember why I put this complex check.

ai:
You asked really good question! Unfortunately, I think we should fix Browserslist, because this error still will be in other tools which use Browserslist.

Do you want to send PR to Browserslist to save your name in the history?

ken1987:
wow ~ I can try

ai:
Awesome. I am closing this PR and will wait for new PR in Browserslist. Thanks for finding this issue, good work.

PR #981, opened 2018-02-03
Don't require sub-files in postcss

### LINKED ISSUE #980: Don't read internal files from postcss
Currently autoprefixer does things like [`const list = require('postcss/lib/list');`](https://github.com/postcss/autoprefixer/blob/master/lib/hacks/cross-fade.js#L2) 
instead of `const list = require('postcss').list;`

This makes bundling autoprefixer, postcss, and other postcss plugins with bundle splitting very difficult to do without duplicating parts of postcss.

Additionally, it could be argued this is breaking [point 2.4 the plugin guidelines](https://github.com/postcss/postcss/blob/master/docs/guidelines/plugin.md#24-use-only-the-public-postcss-api), since the paths aren't documented in the API docs. This prevents postcss from changing its internal directory structure.

Would you be open to changing this?

RyanZim:
Fixes #980

### DIFF
--- a/lib/hacks/cross-fade.js
+++ b/lib/hacks/cross-fade.js
@@ -1,5 +1,5 @@
 const Value = require('../value');
-const list = require('postcss/lib/list');
+const list = require('postcss').list;
 
 class CrossFade extends Value {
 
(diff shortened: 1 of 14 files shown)


ai:
I release it with Browserslist 3 on next week.

RyanZim:
Thanks!

PR #991, opened 2018-02-22
Readme stale counts removed, wording improved

nicnilov:
Readme uses absolute counts which seem to be stale. `Features` section lists 3 options instead of 4 and the `Options` sections lists 9 instead of 8.

Absolute counts removed and the wording improved in terms of grammar.


ai:
Good idea :+1:

Issue #1001, opened 2018-03-12
-webkit-print-color-adjust


Malvoz:
Consider adding `-webkit-print-color-adjust` [(MDN Reference)](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-print-color-adjust) as prefix for `color-adjust` [(CSS Color Module Level 4 Draft)](https://drafts.csswg.org/css-color-4/#color-adjust)

Both properties keyword values are: `economy | exact`.

There is a [data support suggestion on caniuse](https://github.com/Fyrd/caniuse/issues/2634).

ai:
@Malvoz good idea. Do you want to send PR?

Malvoz:
@ai Although I've been around on Github for a while I'm still not sure how to go about PR's.. What are the implications of a PR?

ai:
@Malvoz oops, sorry :).

PR = pull request.

1. Click Fork button in [Autoprefixer repository page](https://github.com/postcss/autoprefixer).
2. It will create your own Autoprefixer’s fork.
3. Clone it to your computer by git.
4. All `yarn install` or `npm install`.
5. Change `data/prefixes.js`.
6. Create git branch `git checkout -b print-color-adjust`
7. Make a commit to the branch.
8. Push to your fork in GitHub
9. Open your fork on GitHub, you will see button to send pull request.

Malvoz:
@ai 
Thanks, I appreciate the instructions, I guess it's time to stop looking the other away and learn git.. I might find time for it tonight, if this PR can wait? 😆

ai:
Sure. Anyways I will have time for it only in Friday.

Malvoz:
@ai 

I've been busy and I wont be able to priorities learning git within the next couple of days. 

I understand this probably shouldn't wait for too long (although probably not a wide usage of this property yet) and because I cannot give you a reliable time frame - if you or anyone else want to PR this then by all means go ahead. 

I'll make sure to get on with git until the next contribution I make!

ai:
@Malvoz don’t worry, I will send developer from Cult of Martians. I plan they will close this issue in this week.

soul-wish:
@Malvoz @ai don't you mind if I'll try to work on this issue?

ai:
@soul-wish sure!

soul-wish:
@ai great! Currently working on a PR to caniuse database.

YozhikM:
Ready hack and tests, we are waiting for the approval of the PR

[...]

PR #1009, opened 2018-03-20
Add color-adjust declaration support

soul-wish:
Here we go in fixing #1001 issue. @ai, can you check this PR, please? Because I'm not sure about some modifications I've made. Haven't found documentation or contribution guide for Autoprefixer, unfortunately. What do you think about adding such type of documentation/guide in the future? Even simple `CONTRIBUTING.md` file would be helpful for newcomers.
I'll try to leave some comments in my commit below ;)

### DIFF
--- a/data/prefixes.js
+++ b/data/prefixes.js
@@ -779,3 +779,12 @@ f(over, { match: /a #1/ }, browsers =>
         browsers
     })
 );
+
+// CSS color-adjust property
+f(require('caniuse-lite/data/features/css-color-adjust.json'), browsers =>
+    prefix(['color-adjust'], {
+        mistakes: ['-khtml-', '-ms-', '-moz-', '-o-'],
+        feature: 'color-adjust',
+        browsers
+    })
+);
--- a/data/prefixes.js
+++ b/data/prefixes.js
@@ -780,7 +780,7 @@ f(over, { match: /a #1/ }, browsers =>
     })
 );
 
-// CSS color-adjust
+// color-adjust
 f(require('caniuse-lite/data/features/css-color-adjust.js'), browsers =>
     prefix(['color-adjust'], {
         feature: 'css-color-adjust',
--- a/data/prefixes.js
+++ b/data/prefixes.js
@@ -780,10 +780,10 @@ f(over, { match: /a #1/ }, browsers =>
     })
 );
 
-// CSS color-adjust property
-f(require('caniuse-lite/data/features/css-color-adjust.json'), browsers =>
+// CSS color-adjust
+f(require('caniuse-lite/data/features/css-color-adjust.js'), browsers =>
     prefix(['color-adjust'], {
-        feature: 'color-adjust',
+        feature: 'css-color-adjust',
         browsers
     })
 );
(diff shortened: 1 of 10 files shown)


data/prefixes.js
```diff
@@ -779,3 +779,12 @@ f(over, { match: /a #1/ }, browsers =>
         browsers
     })
 );
+
+// CSS color-adjust property
+f(require('caniuse-lite/data/features/css-color-adjust.json'), browsers =>
+    prefix(['color-adjust'], {
+        mistakes: ['-khtml-', '-ms-', '-moz-', '-o-'],
```
soul-wish: If I got it in the correct way, this array prevents Autoprefixer from adding unnecessary prefixes. In our case there is no need to add any prefixes except `-webkit-`.

test/autoprefixer.test.js
```diff
@@ -144,7 +144,7 @@ const COMMONS = [
     'advanced-filter', 'element', 'image-set', 'image-rendering',
     'mask-border', 'writing-mode', 'cross-fade', 'gradient-fix',
     'text-emphasis-position', 'grid', 'grid-area', 'grid-template',
-    'grid-template-areas'
+    'grid-template-areas', 'color-adjust'
```
soul-wish: Should I add 'color-adjust' declaration to the array of `COMMONS`? Not sure about this change. It was a little hard to get the whole structure and logic inside this test file.

test/autoprefixer.test.js
```diff
@@ -422,6 +422,7 @@ describe('hacks', () => {
     it('supports text-decoration',      () => test('text-decoration'));
     it('ignores modern direction',      () => test('animation'));
     it('supports overscroll-behavior',  () => test('overscroll-behavior'));
+    it('supports webkit prefixes',      () => test('color-adjust'));
```
soul-wish: Is the name of test is ok? I tried to get the correct naming logic from the test cases above.

ai: Nope, the test name is incorrect. Check others, they contain hack name, not prefix.

test/autoprefixer.test.js
ai: Yes, you should add it to `COMMONS` to run `color-adjust` with not-duplication and prefixes removing tests

data/prefixes.js
ai: Not, it is not correct. Check other prefixes, they don’t have `mistakes`. `mistakes` is only for popular mistakes. I think nobody wrote `-khtml-color-adjust` in the real world.

soul-wish:
@ai thanks for all your help! I've updated PR according to your instructions. Let me know if I've missed something.

soul-wish:
@ai, @yuriyalekseyev found some issues in original PR, so there is updated version. 😉

yuriyalekseyev:
@soul-wish @ai  it seems that tests are green with a new version of caniuse 👍

soul-wish:
Now we have issue only with size-limit package and increased package size. I hope @ai will tell us what do do with this issue (not sure we need to modify size-limit configuration inside our PR).

Issue #1016, opened 2018-04-10
Not everything is translated for ms from area to columns / rows


shiroganee:
I've scss as below

 ```
   &__header {
        display: grid;
        grid-template-columns: auto 380px;
        grid-template-areas: "header header" "carousel info";

        @media (min-width: $screen-sm-min) and (max-width: $screen-md-max) {
            grid-template-columns: auto 310px;
        }

        @media (max-width: $screen-xs-max) {
            grid-template-areas: "header" "carousel" "info";
            grid-template-columns: auto;
        }

        &--variants {
            grid-template-areas: "header info" "carousel info";
        }
    }
```

Instead of creating some two different spans / ends depending on the screen I only get one with the last declared (header info carousel info).

How to make it that I will get at least few different configurations for each area declared?

ai:
Duplicate for https://github.com/postcss/autoprefixer/issues/954

/cc @evgeny-petukhov

Issue #954, opened 2017-12-13
CSS-Grid wrong ms prop values w/ media queries


pcjmfranken:
**Description**
When a container with `display: grid;` changes the `grid-template-areas` property in a media query, children with a `grid-area` property will end up with values for `-ms-grid-row`, `-ms-grid-column` and `-ms-grid-column-span` that correspond to whichever `grid-template-areas` comes first, any other instances are ignored resulting in wrong placement.

**Expected or desired result**
*Depending on whether you consider this a bug, an oversight or a feature request ;)*

Children with a `grid-area` property should be rendered once inside a new media query for each time its parent's `grid` changes with MS values corresponding to these new grid values.

**Example**
I don't know how to add the `grid: true` config option to CodePen's Autoprefixer so unfortunately, you're going to have to try this example locally. [Here's](https://codepen.io/pcjmfranken/pen/rpaeyM) a link to the pen anyway, in case anyone knows.

If you let Autoprefixer transpile the following code you'll notice that the third box will have only received -ms- property values for the 768px media query and that it will still span two columns and be on a second row which no long exists above 1200px.

```html
<div class="container">
  <div class="box boxFirst">First box</div>
  <div class="box boxSecond">Second box</div>
  <div class="box boxThird">Third box</div>
</div>
```

```css
@media screen and (min-width: 768px) {
  .container {
    display: grid;
    grid-template-columns: 1fr 1fr;
    grid-template-rows: auto;
    grid-template-areas:
      "boxFirst boxSecond"
      "boxThird boxThird";
  }
}

@media screen and (min-width: 1200px) {
  .container {
    grid-template-columns: 1fr 1fr 1fr;
    grid-template-rows: auto;
    grid-template-areas:
      "boxFirst boxSecond boxThird";
  }
}

.box { padding: 16px; text-align: center; color: white; font-family: sans-serif; }

.boxFirst { grid-area: boxFirst; background-color: #90afc5;  }
.boxSecond { grid-area: boxSecond; background-color: #2a3132; }
.boxThird { grid-area: boxThird; background-color: #763626; }
```

After transpilation:
```css
.boxThird {
  background-color: background-color: #763626;
  -ms-grid-row: 2;
  -ms-grid-column: 1;
  -ms-grid-column-span: 2;
  grid-area: boxThird;
}
```

Expected after transpilation:
```css
@media screen and (min-width: 768px) {
  .boxThird {
    background-color: background-color: #763626;
    -ms-grid-row: 2;
[...]

PR #1018, opened 2018-04-11
[grid-template-areas] Supported grid-template(-areas) in media rules

evgeny-petukhov:
Problem is mentioned in the issues #954 #1016 and discussed in the issue #954

### DIFF
--- a/test/cases/grid-template-areas.out.css
+++ b/test/cases/grid-template-areas.out.css
@@ -32,32 +32,20 @@
     grid-area: foot;
 }
 
-.f {
-    -ms-grid-row: 3;
-    -ms-grid-column: 2;
-    grid-area: foot;
-}
-
 @media (max-width: 1000px) {
     .a {
         grid-template-areas: "head main";
     }
-
     .b {
-        -ms-grid-row: 1;
-        -ms-grid-column: 1;
-        grid-area: head;
+        -ms-grid-row:
+        1;
+        -ms-grid-column:
+        1;
     }
-
     .c {
-        -ms-grid-row: 1;
-        -ms-grid-column: 2;
-        grid-area: main;
-    }
-
-    .d {
-        -ms-grid-row: 1;
-        -ms-grid-column: 2;
-        grid-area: main;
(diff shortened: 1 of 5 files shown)


test/cases/grid-template-areas.out.css
```diff
@@ -32,32 +32,20 @@
...
     .a {
         grid-template-areas: "head main";
     }
-
     .b {
-        -ms-grid-row: 1;
-        -ms-grid-column: 1;
-        grid-area: head;
+        -ms-grid-row:
+        1;
```
evgeny-petukhov: There is a little problem here with whitespaces. Is it needed to merge raws from different rules/decls for the better result?

ai: Just trim spaces around value and remove `raws.between`

evgeny-petukhov: Used `raws.between` of the origin declaration

test/cases/grid-template.out.css
```diff
@@ -91,3 +91,34 @@
     -ms-grid-column-span: 9;
     grid-area: ft;
 }
+
+@media (min-width: 1000px) {
+    .b {
+        -ms-grid-rows:
+            30px auto 1fr 30px;
```
ai: I think we should not have new line here and l in next property. Let's clean `between` here too

evgeny-petukhov: It's a little different case. These declarations extends `between` from origin which already had new line. So it's better to control further fixes with user's styleguide or linter. But I can fix test and remove new line from origin.

evgeny-petukhov: @ai What do you think?

evgeny-petukhov: @ai About my second commet

ai: Just remove `between`. Do not copy it from an origin. PostCSS will fill it automatically with own space detect algorithm.

ai:
Released in 8.3.

Issue #1024, opened 2018-04-23
[Feature Request] Control Comments PerLine


levin-du:
Control Comments per line or multi-lines is very useful just like eslint does.  Thanks.

ai:
Yeap. Good idea.

ai:
Do you want to send PR?

I will have to implement it only in June.

levin-du:
Interested in this question. Would you like to give me some hits on how to implement? It seems have some test case already.

ai:
I think you should add some code here https://github.com/postcss/autoprefixer/blob/master/lib/processor.js#L340

LonelyGriffin:
I have question. How do the 'ignore next' comment should  interact with nested the 'on' comment? Do it should ignore whole rule in spite off the 'on' comment or conversely.

ai:
@LonelyGriffin I think ignore should disable adding a prefix to a selector or to at-rule, but ignore will not preventing adding prefixes to children. So no problems with nested `on` :D

ai:
Done https://github.com/postcss/autoprefixer/pull/1027

ai:
Released in 8.4

ai:
@simevidas I agree with you. We even have this text in the docs.

But since they anyway need to find a docs to find this feature, it is OK to have it. We had a disabling comments. And a lot of users understand it wrong. So now we just fixed old control comments. Not add them.

simevidas:
@ai By docs, do you mean the section on control comments in the README [1]? That section does not provide usage guidelines; the provided example is `transition: 1s;`. Meanwhile, the 8.4 release notes [2] provide the examples `user-select: none;` and `::placholder`.  

Is there a legitimate reason to disable prefixing for any of these three examples? If not, doesn’t this mean that these are bad examples?

However, if there *are* good reasons to disable Autoprefixer in these cases, I would recommend that these reasons are explained via a note. Some explicit guidelines are definitely needed (or even a warning).

[1]: https://github.com/postcss/autoprefixer#control-comments
[2]: https://github.com/postcss/autoprefixer/releases/tag/8.4.0

ai:
@simevidas PR is welcome

PR #1027, opened 2018-04-27
"autoprefixer: ignore next" support

vostrik:
#1024 Added support for comment "autoprefixer: ignore next"

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -349,10 +349,12 @@ class Processor {
                 /(!\s*)?autoprefixer:\s*ignore\s+next/i
                     .test(previousNode.text)) {
                 node._autoprefixerDisabled = true;
+                node._autoprefixerOnlySelfDisabled = true;
                 return node._autoprefixerDisabled;
             }
         }
 
+        let value = null;
         if (node.nodes) {
             let status = undefined;
             node.each(i => {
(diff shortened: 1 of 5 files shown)


test/cases/ignore-next.out.css
```diff
@@ -0,0 +1,29 @@
...
+  mask: none;
+  background: -webkit-linear-gradient(top, white, black);
+  background: -o-linear-gradient(top, white, black);
+  background: linear-gradient(to bottom, white, black);
+}
+
+/* autoprefixer: ignore next */
+ingore-rule {
+  mask: none;
+  background: linear-gradient(to bottom, white, black);
```
ai: We need prefixes here and inside next at-rule. This comment should affect only selector and at-rule params on next node.

Or we will need to think about combining `ignore next` and `off/on`.

lib/processor.js
```diff
@@ -342,6 +342,18 @@ class Processor {
         if (node._autoprefixerDisabled !== undefined) {
             return node._autoprefixerDisabled;
         }
+        if (node.parent && node.parent.nodes) {
+            const nodeIndex = node.parent.nodes.indexOf(node);
+            if (nodeIndex > 0) {
+                const previousNode = node.parent.nodes[nodeIndex - 1];
```
ai: You can just call `Node#prev` shortcut

vostrik: Fixed

test/cases/ignore-next.out.css
vostrik: Fixed.  Disable Autoprefixer only for selector, at-rule or property.

PR #1028, opened 2018-04-28
fix Processor.disabled(): check if node has parent before get prev()

Diablohu:
If not do so, there are fatal errors, at least on my projects...

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -345,11 +345,13 @@ class Processor {
             return node._autoprefixerDisabled;
         }
 
-        const prev = node.prev();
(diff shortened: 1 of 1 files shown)


ai:
Strange. This issue was fixed in latest PostCSS release 6.0.22.

ai:
@Diablohu you got this error after `npm update`?

ai:
Released in 8.4.1

Diablohu:
I now know what cause this. You are right that PostCSS v6.0.22 fixed this. But `postcss-loader` dose not update yet...

Issue #951, opened 2017-12-07
Add support for grid-gap


pcjmfranken:
The `grid-column-gap`, `grid-row-gap` and their shorthand `grid-gap` ([spec](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-gap)) allow you to set gutters on the inside of the grid.

It would be nice if this property could be transformed as well, perhaps by automatically adding empty columns/rows in between the provided tracks with a size of whatever is provided by the `grid-gap` property.

ai:
@pcjmfranken Do you know any `-ms-` property or another way to emulate `grid-gap` on IE?

pcjmfranken:
There is no -ms- prefix. In order to achieve this transformation, autoprefixer will have to remove the user-provided grid-gap properties and place empty grid tracks between all cells.

[Codepen example here](https://codepen.io/rachelandrew/pen/jVwPdO) (found on [through](https://rachelandrew.co.uk/archives/2016/11/26/should-i-try-to-use-the-ie-implementation-of-css-grid-layout/) page).

ai:
Oh. It looks like too complex. I am not sure that I will be able to find somebody to do this work.

What about use `padding` and don't use `grid-gap`?

pcjmfranken:
It is indeed a bit complex and your alternative solution works.

It would be a nice addition to autoprefixer's css-grid support if someone can find the time to implement this, but it's a feature we can absolutely live without.

ai:
Close since there was no ideas about how to make polyfill

evgeny-petukhov:
I've been working on `grid-gap`. It can be implemented when all these properties are used: `grid-template-areas`, `grid-template-columns`, `grid-template-rows`, `grid-gap`.
Now it's the only case when we know in advance all cells of the grid, and so can position correctly all cells taking into account gutters.

evgeny-petukhov:
@ai I think it's time to reopen this issue for a while.

PascalAOMS:
Why is the way of adding more `-ms-grid-columns` in the size of the given `grid-gap` no option?

Why would we need `grid-template-areas`?

seaneking:
🎉 

@evgeny-petukhov does your shim take into account only setting `grid-collumn-gap` (and therefore only `grid-template-columns`)? 

That's a common use-case IMO, especially for the old standard X-col full page grid. Assume it would just be a subset of the logic you're using to shim both row and column gaps?

[...]

PR #1032, opened 2018-05-02
Add support for grid-gap

evgeny-petukhov:
#951

### DIFF
--- a/lib/hacks/grid-rows-columns.js
+++ b/lib/hacks/grid-rows-columns.js
@@ -1,5 +1,8 @@
 const Declaration = require('../declaration');
-const utils = require('./grid-utils');
+const {
+    prefixTrackProp,
+    prefixTrackValue
+} = require('./grid-utils');
 
 class GridRowsColumns extends Declaration {
 
@@ -13,7 +16,7 @@ class GridRowsColumns extends Declaration {
      */
     prefixed(prop, prefix) {
         if (prefix === '-ms-') {
-            return prefix + prop.replace('template-', '');
+            return prefixTrackProp({ prop, prefix });
         } else {
             return super.prefixed(prop, prefix);
         }
@@ -31,7 +34,7 @@ class GridRowsColumns extends Declaration {
      */
     set(decl, prefix) {
         if (prefix === '-ms-' && decl.value.indexOf('repeat(') !== -1) {
-            decl.value = utils.changeRepeat(decl.value);
+            decl.value = prefixTrackValue({ value: decl.value });
         }
         return super.set(decl, prefix);
     }
--- a/lib/hacks/grid-template-areas.js
+++ b/lib/hacks/grid-template-areas.js
@@ -1,12 +1,18 @@
 const Declaration = require('../declaration');
 const {
     parseGridAreas,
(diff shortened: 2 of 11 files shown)


ai:
Awesome.

What do you think if I asked on Twitter, is it a safe way to work with `grid-gap`?

evgeny-petukhov:
I think it's great idea to ask community. Do you want to receive meanings of experts, or do you want to know that people want to use it?

ai:
I think feedback depends on ideas, not on author. Many experts are bad in some questions. Many unknown person are good 😄

evgeny-petukhov:
Agreed with you, so that is why I asked this question. Let's discuss!

evgeny-petukhov:
@ai Please, notify me about the tweet. I would like to join the discussion 😀

ai:
Asked https://twitter.com/Autoprefixer/status/992210938441928704

Let’s wait few days and then I will accept and release it.

jhpratt:
What happens in this example?

```css
body {
  display: grid;
  grid-gap: 10px;
  grid-template-rows: 1fr 1fr;
}
div {
  grid-row: span 2;
}
```

Due to the auto-placement of items on the grid and the nature of selectors, the behavior of `div` would change. If it were a direct child of `body`, it would have to be `span 3`, as you need to account for the new "row" (the gap). If it's not a direct child, it would have to stay `span 2`. I suppose you could add another selector for direct childs, but something about that just feels off.

As @ai pointed out on Twitter, `grid-template-areas` would avoid this. The only problem is not everyone uses that.

evgeny-petukhov:
@jhpratt Actually, your example mostly isn't related to PR.
> I want to note that it's required grid-template-areas (!), grid-template-rows and grid-template-columns in the same rule to implement grid-gap. In the other cases, grid-gap is skipped as before https://twitter.com/PetukhovYevgeny/status/992280342932291584

So in your example, `grid-gap` will be skipped, and in IE there will not be gutters at all.
```css
body {
  display: grid;
  grid-gap: 10px;
  grid-template-rows: 1fr 1fr;
}
div {
  grid-row: 1 / span 2; // it's need to specify start
}

/* output */
body {
  display: -ms-grid;
  display: grid;
  grid-gap: 10px;
  -ms-grid-rows: 1fr 1fr;
      grid-template-rows: 1fr 1fr;
}
div {
  -ms-grid-row: 1;
  -ms-grid-row-span: 2;
  grid-row: 1 / span 2;
}
```

petamoriken:
Sorry for my interrupt, `grid-row-gap`, `grid-column-gap`, and `grid-gap` has merged with the prefix (`grid-`) less properties. Please resolve this as well. 😅
https://drafts.csswg.org/css-grid/#change-2016-grid-gap
https://drafts.csswg.org/css-align-3/#gap-legacy

ai:
@petamoriken use `postcss-preset-env`. It is PostCSS plugins preset with many polyfills. Including `gap` → `grid-gap` https://github.com/jonathantneal/postcss-gap-properties

Just set it before Autoprefixer

ai:
Released in 8.5

Issue #1049, opened 2018-05-19
[css-grid] `grid-template` without rows defined outputs empty rows template


Dan503:
This source code:
````css
.grid {
  display: grid;
  grid-gap: 10px;
  grid-template:
  "a  b"      /* <<< note that rows are not explicitly set */
  "c  d" /
  1fr 1fr;
}

.cell-A {
  grid-area: a;
}
...
````
Compiles to
````css
.grid {
  display: -ms-grid;
  display: grid;
  grid-gap: 10px;
  -ms-grid-rows: ; /* <<<<  Note the lack of rows */
  -ms-grid-columns: 1fr 10px 1fr;
      grid-template: "a b"
 "c d"/
 1fr 1fr;
}

.cell-A {
  -ms-grid-row: 1;
  -ms-grid-column: 1;
  grid-area: a;
}
...
````
Expected output:
````css
.grid {
  display: -ms-grid;
  display: grid;
  grid-gap: 10px;
  -ms-grid-rows: auto 10px auto; /* <<<<  rows filled in with `auto` values */
  -ms-grid-columns: 1fr 10px 1fr;
      grid-template: "a b"
 "c d"/
 1fr 1fr;
}

.cell-A {
  -ms-grid-row: 1;
  -ms-grid-column: 1;
  grid-area: a;
}
...
````

yepninja:
Thank you for the good examples. They will be very suitable for tests.

Dan503:
@yepninja this is a high priority issue that should get fixed before the article is published.

yepninja:
Released in 8.5.2.

PR #1052, opened 2018-05-26
Fix grid template parsing in case of auto row sizes

y-gagar1n:
#1049 I added some additional checks so that row size would be defined as 'auto' if not specified.

### DIFF
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -201,8 +201,7 @@ function testTrack(node) {
 }
 
 function verifyRowSize(result) {
-    if (result.prop === 'grid-template' &&
-        result.rowProps && !result.sizedRow) {
+    if (result.rowProps && !result.sizedRow) {
         result.rows.push('auto');
     }
     return result;
@@ -246,7 +245,6 @@ function parseTemplate({
             return result;
         }, {
             key: 'rows',
-            prop: decl.prop,
             rowProps: false,
             sizedRow: false,
             columns: [],
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -201,7 +201,7 @@ function testTrack(node) {
 }
 
 function verifyRowSize(result) {
-    if (result.rowProps && !result.sizedRow) {
+    if (result.areas.length > result.rows.length) {
         result.rows.push('auto');
     }
     return result;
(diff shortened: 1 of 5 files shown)


yepninja:
@y-gagar1n This case is about `grid-gap`, so it's better to put tests into `grid-gap.css`.

Also, you can use simpler logic. After `grid-template` parsed you can check rows. If there are no rows, you can add `auto` values according to areas size.

yepninja:
Now I understand your test cases. So we need at least two tests in different files (without and with gap).

lib/hacks/grid-utils.js
```diff
@@ -200,6 +200,14 @@ function testTrack(node) {
     return node.type === 'word' && /^\[.+\]$/.test(node.value);
 }
 
+function verifyRowSize(result) {
+    if (result.prop === 'grid-template' &&
```
yepninja: It doesn't need to check prop value. It's always `grid-template`.

lib/hacks/grid-utils.js
```diff
@@ -213,22 +221,34 @@ function parseTemplate({
...
+                    result.sizedRow = true;
+                }
             }
 
             // devider(/)
             if (type === 'div' && value === '/') {
                 result.key = 'columns';
+                result = verifyRowSize(result);
+                result.rowProps = false;
             }
```
yepninja: Maybe it is enough to compare the length of areas and rows. And if the length of rows is less then the length of areas it needed to push `auto` to rows array.

yepninja:
Maybe it is enough to compare the length of areas and rows. And if the length of rows is less then the length of areas it needed to push auto to rows array.

ai:
@yepninja we are ready to merge?

y-gagar1n:
@ai @yepninja sorry, I'm not finished yet. Commit placing tests in different files will be ready a bit later.

y-gagar1n:
> Maybe it is enough to compare the length of areas and rows. And if the length of rows is less then the length of areas it needed to push auto to rows array.

I still don't think it's possible to handle the case with mixed sized and unsized rows that way.  Imagine we had two different cases:

```css
.a {
  grid-template:
      "a b"
      "c d" 20px;
}

.b {
  grid-template:
      "a b" 20px
      "c d" ;
}
```

Both of them would output `rows` equal to `20px` and information about which row had this size would be lost completely during the `parseTemplate` function. Appending `auto` will result in `20px auto` which is wrong in case of `.a`.

y-gagar1n:
@ai @yepninja I'm finished now.

lib/hacks/grid-utils.js
```diff
@@ -200,6 +200,13 @@ function testTrack(node) {
     return node.type === 'word' && /^\[.+\]$/.test(node.value);
 }
 
+function verifyRowSize(result) {
+    if (result.rowProps && !result.sizedRow) {
```
yepninja: I mean that you can just compare length here.
```js
if (result.areas.length > result.rows.length) result.rows.push('auto')
```
So you don't need additional properties (`rowProps` and `sizedRowz`)

y-gagar1n: Thanks, now I finally see what you mean. I'll commit later today.

yepninja:
@y-gagar1n Great PR! Thank you for contributing!
@ai Now we are ready to merge.

ai:
Thanks. I will release it tomorrow.

ai:
Released in 8.5.2.

Issue #1048, opened 2018-05-19
[css-grid] move media query cell placements to come after cell styles


Dan503:
Currently if I write this:

````scss
/* How I would like to structure my scss file */

.grid {
  display: grid;
  grid-template-columns: repeat(2, 1fr);
  grid-template-areas:
    "a   b"
    "c   d";

  // Placing media query close to the other .grid code
  @media (min-width: 600px){
    grid-template-areas:
      "a   b   c   d";
    grid-template-columns: repeat(4, 1fr);
  }

  &__cell {
    &--a {
      grid-area: a;
    }
    &--d {
      grid-area: d;
    }
  }
}
````

auto-prefixer outputs this:

````css
/* How Autoprefixer outputs that code */

.grid {
 display: -ms-grid;
 display: grid;
     -ms-grid-columns: (1fr)[2];
     grid-template-columns: repeat(2, 1fr);
     grid-template-areas: "a   b"
"c   d";
}

@media (min-width: 600px) {
 .grid {
       grid-template-areas: "a   b   c   d";
   -ms-grid-columns: (1fr)[4];
   grid-template-columns: repeat(4, 1fr);
 }
 .grid__cell--a {
   -ms-grid-row: 1;
   -ms-grid-column: 1;
 }
 .grid__cell--d {
   -ms-grid-row: 1;
   -ms-grid-column: 4;
 }
}

.grid__cell--a {
 -ms-grid-row: 1;
 -ms-grid-column: 1;
 grid-area: a;
}
.grid__cell--d {
 -ms-grid-row: 2;
 -ms-grid-column: 2;
 grid-area: d;
}
````

The issue is that the media query code is being overwritten by the default styles so the media query ends up having no effect in IE.

You can write the input code like this instead which fixes the issue:

````scss
/* How I have to structure my code to not cause the issue */

.grid {
  display: grid;
  grid-template-columns: repeat(2, 1fr);
  grid-template-areas:
    "a   b"
    "c   d";

  &__cell {
    &--a {
      grid-area: a;
    }
    &--d {
      grid-area: d;
    }
  }

  // Placing media query AFTER the cell styles
  @media (min-width: 600px){
    grid-template-areas:
      "a   b   c   d";
    grid-template-columns: repeat(4, 1fr);
  }
}
````

However it means you have to write half of your `.grid` styles at the top of the scss file and write the other half of the styles at the bottom. It is much nicer having them all centralised in one location. 

[...]

PR #1055, opened 2018-06-06
Fix placements of media query with grid-template-area

YozhikM:
#1048 
cc @ai

### DIFF
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -344,8 +344,8 @@ function insertAreas(areas, decl, result) {
             const root = parentMedia.parent;
             const parentMediaIdx = root.index(parentMedia);
             const gridAreaIdx = root.index(gridArea.parent);
-            const nextRuleIdx = root.nodes[gridAreaIdx + 1];
-            const isSameRule = isSameNextRule(nextRuleIdx, rule.selector);
+            const nextRule = root.nodes[gridAreaIdx + 1];
+            const isSameRule = isSameNextRule(nextRule, rule.selector);
 
             if (gridAreaIdx > parentMediaIdx && !isSameRule) {
                 const newMedia = parentMedia.clone().removeAll();
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -341,7 +341,6 @@ function insertAreas(areas, decl, result) {
                     })
                 ));
 
-
             const root = parentMedia.parent;
             const parentMediaIdx = root.index(parentMedia);
             const gridAreaIdx = root.index(gridArea.parent);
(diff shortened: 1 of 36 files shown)


lib/hacks/grid-utils.js
```diff
@@ -329,8 +341,23 @@ function insertAreas(areas, decl, result) {
                     })
                 ));
 
+
```
ai: double new line

YozhikM:
@ai I want to sleep, I pressed it wrong :)

ai:
@YozhikM crazy fast :D 

@yepninja need your review here

@Dan503 seems like we have a solution for that the last critical issue

Dan503:
Excellent :)

Once it's released I'll test it and update the article. Thanks :)

test/cases/grid-template.css
```diff
@@ -69,10 +69,26 @@
...
-            [footer-left] "foot" 30px [footer-right]
+            [header-left] 'head' 30px [header-right]
+            [nav-left] 'nav' auto [nav-right]
+            [main-left] 'main' 1fr [main-right]
+            [footer-left] 'foot' 30px [footer-right]
             / 1fr;
     }
 }
+
+.foo {
```
yepninja: I think it's better to change existing tests than adding new after all. We already have a large number of tests. And the further the more difficult it becomes to understand them.

ai: I like more tests 😄

ai:
Let's decide about the `lib/` code. I will revert quotes in tests. We need this fix ASAP.

test/cases/grid-template.css
yepninja: And I like more test cases :) In this situation, we duplicate the same test cases many times. We don't increase coverage, just increase the amount of code.

ai: Good point 👍

YozhikM: I suggest leaving only `.az` and `.aa`

test/cases/grid-template-areas.css
```diff
@@ -24,6 +24,10 @@
 
 @media (max-width: 1000px) {
     .a {
-        grid-template-areas: "head main";
+        grid-template-areas: "head main article";
```
YozhikM: I re-used the existed tests

ai:
@YozhikM I was convinced to use single media for all moved areas https://github.com/postcss/autoprefixer/pull/1057

Sorry

Dan503:
What?!? Why?

That's extremely dangerous because it not only moves the grid area code but it also brings all the authors other styles along for the ride. There is a very high risk that this will break the authors styles and then cause resentment amongst the front end community

ai:
@Dan503 single = second

ai:
I mean single for every areas (and one additional for `template-areas`)

Dan503:
Oh right. You were talking about a single duplicate. Not a single media query (ie. Moving the original).

Sorry about the misunderstanding

PR #1056, opened 2018-06-07
remove media query with grid-template-area

An-Tu:


An-Tu:
@ai I saw another PR, but how about this implementation?

yepninja:
@An-Tu We already discussed this solution https://github.com/postcss/autoprefixer/issues/1048#issuecomment-390410907

ai:
@An-Tu sorry, you was late. PRs is to small to merge them. Also I prefer solution with multiple media just after every rule. It generates more CSS, but I think solution is more stable.

Issue #1038, opened 2018-05-10
[css-grid] Report a warning if duplicate area names are detected


Dan503:
Take this css:

````css
.grid-alpha {
  grid-template-areas: "delta  echo";
}

.grid-beta {
  grid-template-areas: "echo  delta";
}

.grid-cell {
  -ms-grid-column: ???; /* what column does .grid-cell go in? */
  arid-area: echo;
}
````

Autoprefixer has no access to the DOM so it doesn't really know what area `.grid-cell` belongs to.

Currently Autoprefixer silently just goes with whatever the last `grid-template-areas` definition was in the style-sheet (in this case `"echo  delta"`). If a user places `.grid-cell` in `.grid-alpha` then it will look great in modern browsers but be placed in the wrong column in IE.

Since it is silent, a user could end up building their whole site blissfully unaware of this issue, using duplicate area names everywhere, then finally open IE to see their whole layout go to shit.

Autoprefixer should detect if there are any area name conflicts and if so, warn users to use unique area names at all times.

evgeny-petukhov:
Good suggestion! I think we can do it.

Dan503:
Oh I just realised that you will need to be careful not to trigger false positives on media queries.

This code would be valid:

````css
.grid {
  display: grid;
  grid-gap: 10px;
  grid-template:
    "a   b" 100px
    "c   d" 100px
    "e   f" 100px /
    1fr  1fr;
}

@media (min-width: 600px){
  .grid {
    grid-template-areas:
      "a   b   c"
      "d   e   f";
    grid-template-columns: repeat(3, 1fr);
    grid-template-rows: repeat(2, 100px);
  }
}

.cell-A {
  grid-area: a;
}

.cell-B {
  grid-area: b;
}

.cell-C {
  grid-area: c;
}

.cell-D {
  grid-area: d;
}
````

yepninja:
@Dan503 `grid-template-areas` in media is processed separately, so it's not a problem.

Dan503:
Not if issue https://github.com/postcss/autoprefixer/issues/1045 is fixed.

I think it would be safe to just ignore area names that are inside media queries and only report a warning if duplicates are found outside of media queries.

ShaggyDude:
+1

ai:
I sent this task to our Cult of Martians.

Dan503:
I'm thinking that if a duplicate area name is found in a media query, if the selector for the element is different then it should trigger the warning.

[...]

PR #1057, opened 2018-06-07
[css-grid] move media query cell placements to come after cell styles

yepninja:
Third solution without duplication media queries.

Now we have solutions for all [3 decisions](https://github.com/postcss/autoprefixer/issues/1048#issuecomment-390410054) that we discussed 😄

### DIFF
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -296,6 +296,9 @@ function insertAreas(areas, decl, result) {
     let missed = Object.keys(areas);
 
     const parentMedia = getParentMedia(decl.parent);
+    const rules = [];
+    const areasLength = Object.keys(areas).length;
+    let areasCount = 0;
 
     decl.root().walkDecls('grid-area', gridArea => {
 
@@ -305,13 +308,6 @@ function insertAreas(areas, decl, result) {
         missed = missed.filter(e => e !== value);
 
         if (area && parentMedia) {
-
-            // skip if grid-template-areas already prefixed in media
-            if (parentMedia.some(
-                i => i.selector === gridArea.parent.selector
-            )) {
-                return undefined;
-            }
 
             // create new rule
             const rule = decl.parent.clone({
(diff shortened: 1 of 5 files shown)


ai:
Oh 😅

I prefer to duplicate medias. It is more stable for the cases (like 2-3 different media).

yepninja:
[Here](https://github.com/postcss/autoprefixer/issues/1048) we decided to implement the solution without duplications.

ai:
Since people afraid of Autoprefixer’s Grids, I think using safest way with a few extra media will be better to promotion.

They will be compressed by gzip and we need max safety.

cc @Dan503

Dan503:
We agreed to do a half/half approach remember?

We don't want to shift author styles around in the declaration order. The goal is to create a single duplicate media query that holds all of the autoprefixer generated area code in it that is placed after the last grid area reference.

See my original example in the issue to see what I mean.

Creating a duplicate query for each area reference creates far too much code bloat and moving the full original media query is extremely dangerous.

Dan503:
Option 2 is the method I'm talking about. That is the one we agreed to.

ai:
@Dan503 but creating a single media for every area seems dangerous for me as well. Code bloat will be fixed by gzip. Are you sure that it is a good reason to not use the safest way?

Dan503:
It's not every single one.

Take a good look at the example code I posted in the issue.

Notice that there is only one duplicate media query. There is not a duplicate for each declaration.

There is basically a duplicate query per media query per grid. It is not a duplicate query per query per area.

ai:
> There is basically a duplicate query per media query per grid. It is not a duplicate query per query per area.

Yeap I got it. And I am afraid that combining different areas (which could be in different place of CSS file) into a that second media query is not safe, since we move areas declaration to a different position.

Dan503:
Yes I know, Autoprefixer can't handle grids sharing the same area names. It has that restriction hard boiled into it's area code which is why I opened issue https://github.com/postcss/autoprefixer/issues/1038

I point this restriction out in the article and explain why it is the case so you won't need to worry about community backlash

ai:
Released in 8.6.1

Dan503:
I'll do the article updates tomorrow. Thanks for getting this over the line :D

PR #1069, opened 2018-06-25
fix(readme): Best Practises -> Best Practices

Splaktar:
according to http://www.dictionary.com/e/practice-vs-practise/ since this is being used as a noun
it should be spelled with a `c` in both American and British English


ai:
Thanks for providing proofs, it was very cute (my English is bad, it is just my typo because of my English level 😅)

Issue #1074, opened 2018-07-02
[css-grid] Grid fallback for IE should require implicit grid item assignment


CyberAP:
IE does not have implicit grids and they can not be safely emulated without knowing markup (assuming markup is always static). That means if there is a `display: -ms-grid;` value without explicit grid areas the grid would fail to display correctly.

Example:

```html
<div class="grid">
   <div>First cell</div>
   <div>Second cell</div>
</div>
```
```css
.grid
{
   display: -ms-grid;
   display: grid;
}
```
The code above would render elements on top of each other in IE, which is not expected.

![Grid fail](https://i.imgur.com/eWbpBvu.png)

ai:
How will modern browsers show this settings-less grid?

ai:
Let’s think for conditions to show a warning.

CyberAP:
That's how it should be and how modern browsers render it:
![Correct behaviour](https://i.imgur.com/1kdPIpM.png)

The problem with this is that it is not fixable at all. Grid items in IE require explicit grid cell assignment.
We can not determine the connected grid items for the particular container. That means that the info message should be shown every time grid fallback is used. 

> Info: grid items must be implicitly assigned to grid cells for IE support

ai:
@CyberAP the core idea of using Autoprefixer is to always use it with `grid-template-areas`

CyberAP:
Setting `grid-template-areas` or `grid-template-columns\rows` won't fix this, unfortunately. Grid items still have to be manually assigned to proper cells and rows. Autoprefixer doesn't inform you about that right now and I think it should (once per process would be fine I guess).

ai:
Autoprefixer will inform you if you explicit ask for auto placement.

If we will have clear conditions, we can inform about other cases as well.

ai:
Of course, we don't tell that any Grid can be converted. This is why Grid feature is behind the flag.

PR #1075, opened 2018-07-02
Add a FAQ section for IE grid support caveats

CyberAP:
See: https://github.com/postcss/autoprefixer/issues/1074

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,27 @@ for prefixing ePub3 properties.
 #### What are the caveats of grid support in Internet explorer?
 
 Internet explorer requires implicit grid item assignment. That means that you have to manually assign your grid items to rows and cells if you want grid support.
+For example if you want a 2 column grid layout you would do this:
+```css
+.grid-container
+{
+  display: grid;
+}
+
+.grid-container > :nth-child(1)
+{
+  grid-column: 1;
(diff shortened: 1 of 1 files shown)


ai:
Good idea. Let's just add free code examples and link to that CSS Trick article.

ai:
I improve this FAQ section 83cbac7

WHat do you think?

CyberAP:
Seems reasonable. But I'd still add an example with no `grid-template*` or `grid-area` set, using only manual grid item assignment. Like this:
```css
.grid
{
  display: grid;
}

.grid > :nth-child(1)
{
  grid-column: 1;
  grid-row: 1;
}

.grid > :nth-child(2)
{
  grid-area: 1 / 2; /* Shorthand for grid-row and grid-column */
}
```

ai:
Why do you like to avoid `grid-template`? It is more readable in my opinion.

CyberAP:
To show that it is not required here and it can work out of the box without it. The only thing that has to be done is grid item assignment.

ai:
In the new section, I cover a little more than just Grid with a missed position. Also, I am not sure that `:nth-child(1)` and `grid-column/row` is the best practice to promote it. It is better to promote BEM selectors and readable position.

PR #1080, opened 2018-07-22
Show warning if grid-area has a non-unique identifier (fixes #1038)

### LINKED ISSUE #1038: [css-grid] Report a warning if duplicate area names are detected

bogdan0083:
This is my first PR which adds warning when grid-area has a non-unique identifier (conflict of names). #1038

Not sure if code is clean enough, but I had a lot of fun working on this issue. Thanks! :)

Some questions:
* In my code I refer to area names in grid-area as "identifiers" ([see this](https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident)). Is it alright?
* How should we handle grid-area inside media? I've omitted checking warnings inside media for now

### DIFF
--- a/lib/hacks/grid-area.js
+++ b/lib/hacks/grid-area.js
@@ -7,7 +7,7 @@ class GridArea extends Declaration {
     /**
      * Translate grid-area to separate -ms- prefixed properties
      */
-    insert (decl, prefix, prefixes, result) {
+    insert (decl, prefix, prefixes) {
       if (prefix !== '-ms-') return super.insert(decl, prefix, prefixes)
 
       const values = utils.parse(decl)
@@ -23,13 +23,6 @@ class GridArea extends Declaration {
       ].forEach(([prop, value]) => {
         utils.insertDecl(decl, prop, value)
       })
-
-      const nonUniqueIdentifiers = utils.findNonUniqueAreaIdentifiers(decl)
-      if (nonUniqueIdentifiers.length > 0) {
-        nonUniqueIdentifiers.forEach(area => {
-          decl.warn(result, `grid-area identifier is not unique: ${ area }`)
-        })
-      }
 
       return undefined
     }
--- a/lib/hacks/grid-area.js
+++ b/lib/hacks/grid-area.js
@@ -7,7 +7,7 @@ class GridArea extends Declaration {
     /**
      * Translate grid-area to separate -ms- prefixed properties
      */
-    insert (decl, prefix, prefixes) {
+    insert (decl, prefix, prefixes, result) {
       if (prefix !== '-ms-') return super.insert(decl, prefix, prefixes)
 
       const values = utils.parse(decl)
@@ -23,6 +23,13 @@ class GridArea extends Declaration {
       ].forEach(([prop, value]) => {
         utils.insertDecl(decl, prop, value)
       })
+
+      const nonUniqueIdentifiers = utils.findNonUniqueAreaIdentifiers(decl)
+      if (nonUniqueIdentifiers.length > 0) {
+        nonUniqueIdentifiers.forEach(area => {
+          decl.warn(result, `grid-area identifier is not unique: ${ area }`)
+        })
+      }
 
       return undefined
     }
--- a/lib/hacks/grid-template-areas.js
+++ b/lib/hacks/grid-template-areas.js
@@ -5,7 +5,8 @@ const {
   prefixTrackProp,
   prefixTrackValue,
   getGridGap,
-  warnGridGap
+  warnGridGap,
+  warnDuplicateNames
 } = require('./grid-utils')
 
 function getGridRows (tpl) {
@@ -69,6 +70,9 @@ class GridTemplateAreas extends Declaration {
         decl,
         result
       })
+
+      // warn if grid-template-areas has a duplicate area name
+      warnDuplicateNames({ decl, result })
 
       const areas = parseGridAreas({
         rows: gridRows,
--- a/lib/hacks/grid-template.js
+++ b/lib/hacks/grid-template.js
@@ -3,7 +3,8 @@ const {
   parseTemplate,
   insertAreas,
   getGridGap,
-  warnGridGap
+  warnGridGap,
+  warnDuplicateNames
 } = require('./grid-utils')
 
 class GridTemplate extends Declaration {
@@ -41,6 +42,9 @@ class GridTemplate extends Declaration {
         decl,
         result
       })
+
+      // warn if grid-template has a duplicate area name
+      warnDuplicateNames({ decl, result })
 
       if ((hasRows && hasColumns) || hasAreas) {
         decl.cloneBefore({
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -1,5 +1,4 @@
 const parser = require('postcss-value-parser')
-const regexp = require('../utils').regexp
 const list = require('postcss').list
 
 function convert (value) {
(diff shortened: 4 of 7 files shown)


ai:
Hi. The code is great.

But `grid-template.css` doesn’t have name conflicts ;).

We must find when different tags use the same name in `grid-area`, not when different parents use the same area in `grid-template`.

But the fix should be fast :).

ai:
The code is great :+1: 

@Dan503 @yepninja what do you think?

ai:
@bogdan0083 why it doesn’t see name conflict between `.c` and `.hd`?

Dan503:
I'll take a close look at this tonight to see if it works how I expect it to.

ai:
Yeap, it is my mistake :(. There are no problems right now. Great work.

Let’s wait @Dan503 review (anyway I want to wait for #1081 before releasing it).

Dan503:
... ummm... ok I'm stumped.

How exactly do I go about testing this pull request? 😕

bogdan0083:
Whoops. I've just pushed a commit. Did I break something?

Dan503:
No, (or well I don't know) I just don't know how to test :P

bogdan0083:
Alright. In the last commit I've added warning for a grid-area even inside media rules. It should trigger inside a media in the following cases: 

* If we're can't find the same selector outside media
* If we found the same selector, but outside grid-area value is different

In all other cases a warning will not be shown.

Dan503:
I've taken a look at your test cases file. It is missing media query tests. It is also missing comma separated list tests.

````css
/* Your current tests */
.g {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "g   g" 100px
        "g   g" 100px
        "h   h" 100px /
        1fr  1fr;
}

.g-conflict {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "g   g" 100px
        "g   g" 100px
        "h   h" 100px /
        1fr  1fr;
}

@media (min-width: 600px) {
  /* This should *not* trigger a warning */
  .g {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "g   h" 100px /
        1fr  1fr;
  }

  /* This *should* trigger a warning */
  .g-conflict-2 {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "g   h" 100px /
        1fr  1fr;
  }
}

/* comma list tests */
/* None of these should throw any warnings (unless specified) */
.i, .j {
  display: grid;
  grid-gap: 10px;
  grid-template:
      "i    j" 100px /
      1fr  1fr;
}

@media (max-width: 600px) {
  .i {
    grid-template:
      "i" 100px
      "j" 100px /
      1fr;
  }
  .j {
    grid-template:
      "i" 100px
      "j" 100px /
      1fr;
  }

  /* This one should throw a warning */
  .k {
    grid-template:
      "i" 100px
      "j" 100px /
      1fr;
  }
}

@media (min-width: 900px) {
  .i, .j {
    display: grid;
    grid-gap: 10px;
    grid-template:
      "i" 100px
      "j" 100px /
      1fr;
  }
}

/* media query test */
@media (min-width: 601px) {
  .l {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "l   m" 100px /
        1fr  1fr;
  }
}
@media (max-width: 600px) {
  .l {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "l   m" 100px /
        1fr  1fr;
  }

  /* this should display a warning */
  .z {
    display: grid;
    grid-gap: 10px;
    grid-template:
        "m   z" 100px /
        1fr  1fr;
  }
}
````

That's all the tests I can think of.

**Comment by bogdan0083** [comment]
Alright, thanks! Looks like I choose a completely wrong approach. I triggered warning only when grid-area is used (which is not very good as I understood now). 

Now I'll just take your code above as test case and show warnings every time there are duplicate area names 👍

**Comment by Dan503** [comment]
> I triggered warning only when grid-area is used  (which is not very good as I understood now).

Having `display: grid;` in the same rule should not be a requirement for the warning to not go off. It should still be powered pretty much entirely through the `grid-template-areas` property if possible.

**Comment by Dan503** [comment]
Yeah the media query and comma seperated list support is what makes this issue much more difficult to fix than it appears to be on the surface.

**Comment by bogdan0083** [comment]
Alright, looks like now it should work as expected inside media rules now 😄 

@Dan503 Can you please check the last commit in your spare time please? These are the warnings I have from the last grid-template.css:
```
"autoprefixer: <css input>:41:5: duplicate area names: head, nav, main, foot",
"autoprefixer: <css input>:133:5: duplicate area names: g, h",
"autoprefixer: <css input>:154:5: duplicate area names: g, h",
"autoprefixer: <css input>:186:5: duplicate area names: i, j",
"autoprefixer: <css input>:228:5: duplicate area name: m",
``` 

And then we can go for a code review! :)

**Comment by Dan503** [comment]
If those are the warnings that those test are giving off now then that sounds like it's working how it's supposed to :)

I feel like the warning message isn't detailed enough though. It warns that duplicate area names are used but it doesn't tell users why duplicate area names are bad or what CSS selector is causing the issue.

If the warning could say something more like this then I think it would be better:

````
autoprefixer: <css input>:41:5:
duplicate area names detected in rule: .g-conflict
duplicate area names: g, h
Duplicate area names cause unexpected behavior in IE
````

**Comment by Dan503** [comment]
You can use `\n` to create the line breaks, or template literals if ES6 is an option.

**Comment by bogdan0083** [comment]
Alright, done 👍 

@ai Now we can review the code I guess (or we can wait until #1081 is fixed, doesn't matter). I changed the implementation completely compared to the initial commit.

**Comment by Dan503** [comment]
I just took a look at the commit. You are using one big template literal which is ok but you need to factor in prefered output white space rather than whatever whitespace lines up nicely with the rest of the code in the file.

The warning will currently be outputting this:
````
autoprefixer: <css input>:133:5: 
        duplicate area names detected in rule: .g-conflict
        duplicate area names: g, h
        duplicate area names cause unexpected behavior in IE
````

A bit of an indent is ok but 8 spaces worth is a bit much :/

A large indent will cause it to go off the screen or wrap in peoples console windows earlier than it needs to.

I'd recommend reducing the indent down to 2 spaces.
````
autoprefixer: <css input>:133:5: 
  duplicate area names detected in rule: .g-conflict
  duplicate area names: g, h
  duplicate area names cause unexpected behavior in IE
````

**Comment by bogdan0083** [comment]
> I just took a look at the commit. You are using one big template literal which is ok but you need to factor in prefered output white space rather than whatever whitespace lines up nicely with the rest of the code in the file.

I've been aware of this issue and I thought maybe someone would give me an advice on fixing the indentation (but forgot to ask).

In the code I use a postcss ```warn``` method to show a warning:
```js
      decl.warn(
        result,
        `
        duplicate area ${ word } detected in rule: ${ rule.selector }
        duplicate area ${ word }: ${ duplicates.join(', ') }
        duplicate area names cause unexpected behavior in IE`
      )
```

the part ```autoprefixer: <css input>:133:5:``` adds automatically before the string.

If I want to indent warning normally, I have to format a warning method like this:
```js
       decl.warn(
        result,
        `
     duplicate area ${ word } detected in rule: ${ rule.selector }
     duplicate area ${ word }: ${ duplicates.join(', ') }
     duplicate area names cause unexpected behavior in IE`
      ) 
```
which is quite an ugly way to indent code.
And the same for the testing code. 

If I use normal string like ```'string' + '\n'``` it still breaks the indentation. 

I searched every warning inside autorefixer – they all are one-liners. I even checked the testing code for warn method (in postcss repository) – there are no multi-line warnings in there as well.

I guess we need to wait what Andrey says on how to fix this (or someone who knows how to make warnings multi-line).

bogdan0083:
> If you don't want to deal with the ugliness, try this instead. It should produce the output with a 2 space indent. 

Thank you very much! Such a simple solution! :smile:

ai:
Are we ready for merge and everything was fixed in the last commit?

bogdan0083:
@ai yep, I finished working on this PR. If code is fine we can merge.

Dan503:
Yep, I'm happy for it to be merged :)

ai:
Released in 9.0.2

ShaggyDude:
This is amazing work. Thank you all!

Issue #1081, opened 2018-07-22
`background-clip: text` doesn't get expected vendor prefix


jmmarco:
Hello :wave:
I was trying to get the necessary vendor prefix for the `background-clip` property using `https://autoprefixer.github.io/`.

Input:
```
p {
  background: coral;
  background-clip: text;
  color: transparent;
}
```

Expected Output:
```
p {
  background: coral;
  background-clip: text;
  -webkit-background-clip: text;
  color: transparent;
}
```
Current Output:
```
p {
  background: coral;
  background-clip: text;
  color: transparent;
}
```

[According to caniuse](https://caniuse.com/#search=background-clip), support for Firefox, Chrome and Safari can be achieved by using the unoficial:
`-webkit-background-clip: text`

ai:
Seems like `text` finally become part of the standard, so we can add support to Autoprefixer.

@jmmarco can you help me and send PR to [`caniuse`](https://github.com/Fyrd/caniuse) with additional JSON file of `text` value? We can’t add support to Autoprefixer without explicit (not comment) data on Can I Use.

ai:
Yeap, you need to create new JSON file about `text` value in `features-json/`

jmmarco:
Done! Let's see if I did well.. [Here's the link to the PR](https://github.com/Fyrd/caniuse/pull/4392)

ai:
@jmmarco awesome! Do you want to send PR to the Autoprefixer as well to got so big project in GitHub profile? You need:

1. Add `text` value with `props: ['background-clip']` data to `data/prefixes.js`
2. Add `lib/hacks/text.js` with the class which will extends `Value`.
3. Load this hack in `lib/prefixes.js`
4. In that hack class change the prefix for IE to `-webkit-`. See [appearance hack](https://github.com/postcss/autoprefixer/blob/master/lib/hacks/appearance.js).
5. Add tests to `autoprefixer.test.js` to test that Autoprefixer with Chrome and Edge browsers adds only `-webkit-` prefix.

HM100:
Edge 15+ Supports the text value without any prefixes according to what I read on MDN before

ai:
@jmmarco yeap @HM100 is right. Can you send additional PR to Can I Use? Also it will be good to add explicit note for Edge, that they used `-webkit-` instead of `-ms-` prefix.

But the plan for Autoprefixer is the same. Autoprefixer anyway need a hack for Edge in case if somebody will need prefixes for Edge 12.

[...]

PR #1086, opened 2018-08-01
[grid-template] fix adding grid span for IE (fixes #1084)

### LINKED ISSUE #1084: Grid Column / Row Span for IE11
The autoprefixer has slight problems with changing the column widths. 

Example a grid with three columns. In the first breakpoint, an element has 2 columns. 

```scss
grid-gap: 10px 20px;
grid-template:
  "log met qu2" 20px
  "log lin lin" auto
  "log nav nav" auto /
  200px 1fr 110px;
```

In the next breakpoint, the element `nav` should now work on a column.

```scss
grid-gap: 10px 50px;
grid-template:
  "log met soc" 20px
  "log lin lin" auto
  "log nav qu2" auto /
  290px 1fr 110px;
```

The problem here is that in the second grid I have `-ms-grid-column-span: 3` still from the previous breakpoint. Which means that the elements now overlap. I have now helped me manually add `-ms-grid-column-span: 1` at the breakpoint.

To automate the whole thing, the autoprefixer should always note `-ms-grid-column-span`, even if the box is only one column wide to overwrite old values in CSS. This also applies to the `-ms-grid-row-span`.

bogdan0083:
This PR fixes #1084.

@Dan503 Can you please review grid-template.out.css in you spare time? I think now it should add -ms-grid-(row|column)-span just fine 👍

### DIFF
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -394,15 +394,11 @@ function getParentMedia (parent) {
 function shouldAddSpan (decl, area, areaName) {
   const root = decl.root()
   const rule = decl.parent
-  const media = getParentMedia(rule)
-  const ruleIndex = root.index(media) + media.index(rule)
   const overrideValues = [false, false]
 
   root.walkRules(rule.selector, node => {
-    const comparedRuleIndex = root.index(node) + media.index(node)
-
     // abort if we are on the same rule
-    if (ruleIndex === comparedRuleIndex) {
+    if (rule.toString() === node.toString()) {
       return false
     }
 
--- a/lib/hacks/grid-utils.js
+++ b/lib/hacks/grid-utils.js
@@ -1,5 +1,5 @@
-const parser = require('postcss-value-parser')
-const list = require('postcss').list
+let parser = require('postcss-value-parser')
+let list = require('postcss').list
 
 function convert (value) {
   if (value &&
@@ -21,15 +21,15 @@ function convert (value) {
 }
 
 function translate (values, startIndex, endIndex) {
-  const startValue = values[startIndex]
-  const endValue = values[endIndex]
+  let startValue = values[startIndex]
+  let endValue = values[endIndex]
 
   if (!startValue) {
     return [false, false]
   }
 
-  const [start, spanStart] = convert(startValue)
-  const [end, spanEnd] = convert(endValue)
+  let [start, spanStart] = convert(startValue)
+  let [end, spanEnd] = convert(endValue)
 
   if (start && !endValue) {
     return [start, false]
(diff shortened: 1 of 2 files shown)


Dan503:
I don't really like the idea of always outputting grid-column/row-span values, but I can also see that trying to pick and choose when to output the span value is going to be a nightmare.

If we go down the always-output-a-span-value route, then I think there is an issue in your pull request. Your span output seems to be inconsistent.

Current output:

```css
    .head {
        -ms-grid-row: 1;
        -ms-grid-column: 1;
        -ms-grid-column-span: 1;
    }
    .nav {
        -ms-grid-row: 2;
        -ms-grid-row-span: 1;
        -ms-grid-column: 1;
    }
```

expected output:
```css
    .head {
        -ms-grid-row: 1;
        -ms-grid-row-span: 1; /* added row span */
        -ms-grid-column: 1;
        -ms-grid-column-span: 1;
    }
    .nav {
        -ms-grid-row: 2;
        -ms-grid-row-span: 1;
        -ms-grid-column: 1;
        -ms-grid-column-span: 1; /* added column span */
    }
```

Basically when grid-areas are being used, Autoprefixer should always output:
- `-ms-grid-row`
- `-ms-grid-row-span`
- `-ms-grid-column`
- `-ms-grid-column-span`

Or is there some smart discrimination going on in your code that is purposely leaving the extra span values out?

bogdan0083:
>  Or is there some smart discrimination going on in your code that is purposely leaving the extra span values out?

Actually there is :). The logic is simple:

If we find grid-template(-area) inside media rule, search the rule above with the same selector and compare -ms-grid-row/column-span values. If values differ then we can add property.  

Before I started working on this issue there was a simple condition that didn't add -ms-grid-row/column-span if one of the values was 1 (e.g ```-ms-grid-row-span: 1```)

> Basically when grid-areas are being used, Autoprefixer should always output:
> -ms-grid-row
> -ms-grid-row-span
> -ms-grid-column
> -ms-grid-column-span

This is very easy to implement actually. The question is: should we add -ms-grid-row/column-span for EVERY rule with grid-template(-areas) or only for the rules inside media rule? (right now it doesn't add property if -ms-grid-row/column-span is 1)

Dan503:
> This is very easy to implement actually.

Nah I prefer keeping the Autoprefixer output as slim as possible. I made that suggestion thinking you might have missed something.

If you have logic to only add the span values when needed then that is a much better solution :)

bogdan0083:
> If you have logic to only add the span values when needed then that is a much better solution :) 

Alright! :smile: So we can now go for a code review I guess :)

Dan503:
You should look into the Travis CLI issues. @ai isn't going to accept the pull request with Travis CLI errors in it.

ai:
Yeap, you need improve tests, so all code should be executed durting test.

Just run `npx jest --coverage` and then check HTML page in `coverage/`

bogdan0083:
@ai Alright, sorry. Now all checks are fine I guess.

ai:
@bogdan0083 thanks! great work

ai:
Sorry, I will take few days to fix this issue https://github.com/postcss/autoprefixer/issues/1081 and release they all together

ai:
Released in 9.1

PR #1111, opened 2018-09-12
Refactoring, delete unnecessary else

janczer:


XhmikosR:
@ai: are you sure this was right? I'm seeing `webkit-box-align` when before I didn't.

ai:
@XhmikosR send PR with a test and reverting this commit

XhmikosR:
@ai I'm not familiar with the codebase, so maybe this wasn't the root cause. I'll make an issue, but going back to 9.1.5 I don't get the `-webkit-box-align` being added.

ai:
@XhmikosR in this case with send PR with a test, which was passed in 9.1.5 and will not be passed in 9.2.

PR #1120, opened 2018-09-18
Adding a basic experimentation playground to the repo

Dan503:
It's been a massive pain trying to test whatever the current build of Autoprefixer is while it's been under development.

I've been resorting to opening the Autoprefixer repo, compiling it, copying the compiled assets into the node_modules folder of a completely unrelated project and then running the compiler for that project to get output CSS.

**That is an insane workflow!**

There needed to be an easier way to play around with the current Autoprefixer build and see if things are working as they are supposed to without having to go to all the trouble of creating formal tests.

So with that in mind I set up this simple little experimentation playground inside the Autoprefixer repo.

1. Run the command `gulp play`
2. Write some CSS in `playground.input.css` and save the file
3. View the output CSS in `playground.output.css`

For some reason it isn't translating CSS Grid code even though I've turned the setting on. @ai Do you have any idea why it isn't outputting CSS Grid code?

### DIFF
--- a/playground/playground.output.css
+++ b/playground/playground.output.css
@@ -1,0 +1,11 @@
+
+/*
+  This file is purely just for experimentation.
+  1. Run the command "gulp play"
+  2. Write some CSS in playground.input.css and save the file
+  3. View the output CSS in playground.output.css
+*/
+
+.grid {
+  display: grid;
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -94,7 +94,7 @@ gulp.task('standalone', ['build:lib'], done => {
 })
 
 gulp.task('compile-playground', () => {
-  let autoprefixer = require('./build')
+  let autoprefixer = require('./build') // @codingStandardsIgnoreLine
   return gulp.src('./playground/playground.input.css')
     .pipe(rename('playground.output.css'))
     .pipe(postcss([autoprefixer({ grid: true })]))
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,8 @@
 let gulp = require('gulp')
 let path = require('path')
 let fs = require('fs-extra')
+let postcss = require('gulp-postcss')
+let rename = require('gulp-rename')
 
 gulp.task('clean', done => {
   fs.remove(path.join(__dirname, 'autoprefixer.js'), () => {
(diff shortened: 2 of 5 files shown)


playground/playground.output.css
```diff
@@ -0,0 +1,11 @@
...
+/*
+  This file is purely just for experimentation.
+  1. Run the command "gulp play"
+  2. Write some CSS in playground.input.css and save the file
+  3. View the output CSS in playground.output.css
+*/
+
+.grid {
+  display: grid;
+}
```
Dan503: This should be outputting:

```css
.grid {
  display: -ms-grid;
  display: grid;
}
```

Dan503:
> 97:30error  "./build" is not foundnode/no-missing-require

Ummm... that is not possible to fix. I've purposely done it that way so that it is running off of the real currently active code.

I purposely did not require it until I could guarrantee that the folder exists.

Can we disable that rule or tell Travis CLI to ignore that line?

Dan503:
[This page on Stack Overflow](https://stackoverflow.com/a/48028726/1611058) says that you can ignore a line of code using this

```
// @codingStandardsIgnoreLine
```

It didn't seem to work though 😥

Dan503:
The alternative is checking the "build" folder into git so that it exists.

ai:
Do not forget to check that `build/` will not contain `playground` after `gulp build`.

ai:
> 97:30error "./build" is not foundnode/no-missing-require

Ignore it (you can fix it by calling `gulp build`)

Dan503:
Ummm... by merging this as it is though, Travis CI will fail for all future pull requests.

I think the easiest way to resolve this is to save build into git.

Dan503:
It's also not outputting grid translations like it's supposed to.

ai:
> I think the easiest way to resolve this is to save build into git.

Nope, generated files should be out of the git.

I just disabled that rule. Anyway it brings me problems in different places.

There is no Grid because `browserslist` in `package.json` has no IE.

Dan503:
Ok, would you prefer the IE support be applied to browserlist in package.json or in the gulp task?

The gulp task will keep it specific to just the playground.

I don't understand why you are excluding IE by default in the browser list 😕

ai:
> I don't understand why you are excluding IE by default in the browser list 😕

I added this Browserslist for Babel. I didn't found when users will need to run Autoprefixer JS files in IE.

> The gulp task will keep it specific to just the playground.

The best way is to put additional Browserslist to `playground/.browserslistrc` with `defaults` query. It will be explicit to users what browsers are used there.

Dan503:
Ok, I'll do that tonight after work 😊👍

Dan503:
> Do not forget to check that `build/` will not contain `playground` after `gulp build`.

Playground doesn't doesn't live in the build folder. It will never be inside that folder.

ai:
> Playground doesn't doesn't live in the build folder. It will never be inside that folder.

Nope, build scripts copy everything to `build/`. But don't worry, I added playground to ignore list.

Dan503:
Don't ignore the whole playground folder. You at least need playground.input.css to stay saved in Git.

Dan503:
playground.output.css is a generated file so it's understandable if you want to ignore that one.

PR #1121, opened 2018-09-19
Playground fixes

Dan503:
This fixes the Gulp watch task and adds CSS Grid translation support


Dan503:
@bogdan0083 I think you will have fun playing with this new toy 😁

1. Run `gulp play`
2. Type CSS into `playground/input.css` and save the file
3. View the output in `playground/output.css`

bogdan0083:
This feature is incredibly awesome! Now we can test branches easily and without any hassle, that's cool!  Thanks @Dan503  :)

Dan503:
Yep! :D

This is a hell of a lot easier than compiling, copying files to a different project node_modules folder and then using that project to test in.

That was an insane work flow.

This is going to make working on the project so much easier :)

PR #1156, opened 2018-11-17
Setting up a VS Code debugger launch.json config file

Dan503:
This makes it immensly easier to debug autoprefixer bugs in VS Code because you can set up break points and have them trigger while running the `gulp play` task.

To take advantage of this, add some breakpoints to one of the core Autoprefixer js files.

VS-Code Break JS point

Now go to the Debug panel and make sure that "Gulp play" is the selected debug option (it should be selected by default).

Debug panel button and the "Gulp play" debug option

Click the green play button.

Green play button

You are now running the `gulp play` task in debug mode. Gulp will function the same as if you typed `gulp play` directly into the terminal with the key exception being that the code will pause execution wherever you have placed breakpoints making bugs immensely easier to debug.

I've also added the default `gulp build` task into the config file as an extra option if you don't want the debugger process to run continuously once started.

(@bogdan0083 Merry Christmas 😜)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,26 @@
+{
+	// Use IntelliSense to learn about possible attributes.
+	// Hover to view descriptions of existing attributes.
+	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"type": "node",
+			"request": "launch",
+			"name": "Gulp play",
+			"program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
+			"args": [
+				"play"
+			]
+		},
+		{
+			"type": "node",
+			"request": "launch",
+			"name": "Gulp build",
+			"program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
+			"args": [
+				"build"
+			]
+		}
+	]
+}


bogdan0083:
Hey hey! 😀 This debug feature is awesome and I've been using it for quite a long time developing features for autoprefixer. I'm certainly for it 🔥

ai:
Do you need to put it to npm package? Or you just want it in your local repository?

ai:
> I'll add it to npmignore.

:+1:

Dan503:
Wait, don't close it! I said **npmignore** not gitignore.

It still needs to be saved into git for everyone to take advantage of the debugging feature, it just doesn't need to downloaded to everyone's computer when they npm install.

ai:
> It still needs to be saved into git for everyone to take advantage of the debugging feature

Not everyone uses VSCode. Not everyone debugs code in your way. It is better just to keep it local and add to `.git/ignore`

Dan503:
>Not everyone uses VSCode. Not everyone debugs code in your way.

It doesn't matter if not _everyone_ debugs using VS Code. What does matter is that for those that _do_ debug using VS Code, this makes it immensely easier for those people. It also means that VS Code developers don't have to set up their own launch.json file to run the debugger because one has already been set up for them.

Dan503:
This does not detriment non-VS-Code devs in any way but it massively helps people who _do_ use VS Code.

ai:
I do not want to add editor-specific files to the project (this is a rule across all my projects). Just because we would have a lot of different files for every editor at the end.

Dan503:
VS Code has a big advantage over other editors for this project thanks to it's debugging features. It is by far the most popular editor for developers right now.

> Just because we would have a lot of different files for every editor at the end.

I don't think that is going to happen.

>I do not want to add editor-specific files to the project (this is a rule across all my projects).

Can you make an exception for VS Code due to how much easier it is to debug on and how popular the editor is among developers?

ai:
I do not see a rational reason for the exception. Nothing will be changed for you if you will keep it locally. We have only a few VS Code developers, both of them can just keep it locally.

Dan503:
Ok, I can see you aren't going to change you're mind about this.

@bogdan0083 if you want to use this for debugging, I've attached a copy of my launch.json file.

[vscode-launch.json.zip](https://github.com/postcss/autoprefixer/files/2592349/vscode-launch.json.zip)

Dan503:
In order to ignore it locally, add a `.vscode` line to the `.git/info/exclude` file.

Issue #1157, opened 2018-11-20
Remove warning from grid-[column/row]-end: span X; + grid-area combo


Dan503:
In #1155 we discovered that `grid-column-end: span 2;` is a valid way to make a grid area cell overlap other areas.

```css
.spanning-tile {
    grid-area: tile-5;
    grid-column-end: span 2; /* force an overlap */
}
```

We should remove the warning for `grid-[column/row]-end: span X;`

>You already have a grid-area declaration present in the rule. You should use either grid-area or grid-column-end, not both

fanich37:
Hi, @Dan503 . Is this ok 8dce36d84cfb68ace6990d0629964d81c2cba8e4?

Dan503:
I'm not going to be able to test this until Sunday night. Thanks for looking into this though :)

Dan503:
@fanich37 can you create a pull request for this? It will make it easier to test and merge.

ai:
Closed by https://github.com/postcss/autoprefixer/pull/1159

PR #1159, opened 2018-11-27
Remove warning from grid-[column/row]-end: span X; + grid-area combo …

fanich37:
…(#1157)

* Add clause to warnIfGridRowColumnExists method in lib/hacks/grid-utils.js

* Update tests


ai:
Awesome. Thanks

PR #1162, opened 2018-12-03
Fixed enabling autoplacement in old projects section recommendation

Dan503:
The previous recommendation of swapping the order around was wrong. The nth-child selector will still overpower the single class selector even if the single class selector comes after the nth-child selector.

If you want, you can just cherry pick the last commit. That is the commit with the Readme fixes in it.

Unfortunately this will need to be released as 9.4.1 if you want the readme fix to appear on npm. ☹

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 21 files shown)


Dan503:
I've created a cleaner pull request here: https://github.com/postcss/autoprefixer/pull/1163

PR #1173, opened 2018-12-24
Gulp 4 code fixes

Dan503:
There were some issues with the Gulp 4 code. This pull request fixes those issues.

- Replaced `path.join()` with a `join()` function (`path.join()` doesn't work properly in Gulp 4)
- It now runs `'build:lib'`, `'build:docs'`, `'build:bin'`, and `'build:package'` all in parallel rather than one after the other for a faster compile time.
- Replaced the old `gulp.start()` commands (not a valid syntax in Gulp 4) with proper `gulp.series()` syntax.
- Fixed the `"watch-playground"` Gulp task syntax.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,12 +1,13 @@
 let gulp = require('gulp')
-let path = require('path')
 let fs = require('fs-extra')
 let postcss = require('gulp-postcss')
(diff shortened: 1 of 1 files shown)


ai:
Thanks 👍

Issue #1152, opened 2018-10-27
`-ms-writing-mode` does not take `direction` property into account


Dan503:
In this code pen, I have both `writing-mode: vertical-rl;` and `direction: rtl;` in the same rule CSS rule.

https://codepen.io/daniel-tonon/pen/rqPbmZ

In modern browsers it looks like this:

![v-rl_d-rtl](https://user-images.githubusercontent.com/10610368/47600215-457a3d80-da09-11e8-926a-50191ee55b98.png)

However in IE it appears like this:

![v-rl_d-ltr](https://user-images.githubusercontent.com/10610368/47600192-e7e5f100-da08-11e8-82e8-06a75485b71f.png)

Usually IE does support `direction` when it is used on it's own however when it is combined with writing mode, `writing-mode` takes priority, even if `direction` was used after `writing-mode` in the CSS order.

**Input CSS**

```css
.element {
  writing-mode: vertical-rl;
  direction: rtl;
}
```

**Current output CSS**

```css
.element {
  -ms-writing-mode: tb-rl;
  writing-mode: vertical-rl;
  direction: rtl;
}
```

**Expected output CSS**
```css
.element {
  -ms-writing-mode: bt-rl;
  writing-mode: vertical-rl;
  direction: rtl;
}
```
----

# Examples of every combination

For the sake of completeness here are a bunch of screenshots and code snippets showing all the different possible configurations you can have:

---

````
direction: ltr;
-ms-writing-mode: lr-tb;
writing-mode: horizontal-tb;
````
![h-tb_d-ltr](https://user-images.githubusercontent.com/10610368/47600103-ccc6b180-da07-11e8-94d2-1512619f35ca.png)

---

````
direction: rtl;
-ms-writing-mode: rl-tb;
writing-mode: horizontal-tb;
````
![h-tb_d-rtl](https://user-images.githubusercontent.com/10610368/47600124-10212000-da08-11e8-80e7-869bb9b5343c.png)

---

````
direction: ltr;
-ms-writing-mode: tb-lr;
writing-mode: vertical-lr;
````
![v-lr_d-ltr](https://user-images.githubusercontent.com/10610368/47600150-5d9d8d00-da08-11e8-8ef1-0692e70596af.png)

---

````
direction: rtl;
-ms-writing-mode: bt-lr;
writing-mode: vertical-lr;
````
[...]

PR #1175, opened 2019-01-01
Add `ms-writing-mode` value based on `direction` declaration

webschik:
Fix for #1152

### DIFF
--- a/lib/hacks/writing-mode.js
+++ b/lib/hacks/writing-mode.js
@@ -4,17 +4,44 @@ class WritingMode extends Declaration {
   static names = ['writing-mode']
 
   static msValues = {
-    'horizontal-tb': 'lr-tb',
-    'vertical-rl': 'tb-rl',
-    'vertical-lr': 'tb-lr'
+    ltr: {
+      'horizontal-tb': 'lr-tb',
+      'vertical-rl': 'tb-rl',
+      'vertical-lr': 'tb-lr'
+    },
+    rtl: {
+      'horizontal-tb': 'rl-tb',
+      'vertical-rl': 'bt-rl',
+      'vertical-lr': 'bt-lr'
+    }
   }
 
-  set (decl, prefix) {
+  insert (decl, prefix, prefixes) {
     if (prefix === '-ms-') {
-      decl.value = WritingMode.msValues[decl.value] || decl.value
-      return super.set(decl, prefix)
+      let cloned = this.set(this.clone(decl), prefix)
+
+      if (this.needCascade(decl)) {
+        cloned.raws.before = this.calcBefore(prefixes, decl, prefix)
+      }
+      let direction = 'ltr'
+
+      decl.parent.nodes.forEach(node => {
+        /**
+         * Skip global values
+         *  direction: inherit;
+         *  direction: initial;
+         *  direction: unset;
+         */
+        if (node.prop === 'direction' && node.value === 'rtl') {
+          direction = node.value
+        }
+      })
+
+      cloned.value = WritingMode.msValues[direction][decl.value] || decl.value
+      return decl.parent.insertBefore(decl, cloned)
     }
-    return super.set(decl, prefix)
+
+    return super.insert(decl, prefix, prefixes)
   }
 }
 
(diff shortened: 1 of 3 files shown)


Dan503:
Thanks for doing this! 😁

One thing I'm worried about is that if IE11 is the only browser that needs to be supported, it won't output the prefixes since caniuse says IE11 doesn't need the -ms- prefixes.

Are you able to ensure that if IE11 is the only supported browser it still outputs the prefixes?

webschik:
Hi @Dan503 !
I use this tool almost in all my projects, personal and at work, so I'm glad to help at least a little :)

Why will it not output the prefixes?

I checked Postcss v7.0.6 and autoprefixer v9.4.0, e.g. in this REPL https://autoprefixer.github.io, with the browser targets `last 2 version` and I've got the prefixed rule:

```
.example {
    writing-mode: vertical-rl;
}

.example {
    -webkit-writing-mode: vertical-rl;
        -ms-writing-mode: tb-rl;
            writing-mode: vertical-rl;
}
```

Caniuse and MDN say that IE has a partial support with depecated values `lr`, `lr-tb`, `rl`, `tb`, `tb-rl` (except for SVG support)

https://caniuse.com/#search=writing
https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode

Dan503:
> it won't output the prefixes since caniuse says IE11 doesn't need the -ms- prefixes.

IE11 doesn't understand the modern syntax but if you use the old syntax unprefixed, it will work in IE11. This means caniuse will say IE11 doesn't need prefixing. Due to it not supporting the modern syntax though, it does need prefixing.

That's what I'm worried about. `last 2 version` includes IE10 so it's not a valid test for this.

Assuming the browser target can be set without breaking other tests, set the browser target to just `ie 11` and check it still outputs prefixes.

webschik:
@Dan503 , I see what you mean.

https://autoprefixer.github.io/, `ie >= 11`

```css
.example {
    writing-mode: vertical-rl;
}

.example {
    -ms-writing-mode: tb-rl;
            writing-mode: vertical-rl;
}
```

I've checked today your test cases from the issue in the real IE11 with the fixed values for `-ms-writing-mode`
and I've got the same result as in Chrome with `writing-mode`. So I see no problem if we use prefixed rule for IE instead of

```
.example {
    writing-mode: tb-rl;
    writing-mode: vertical-rl;
}
```

Dan503:
Also you need to make a test for every iteration posted in the issue and check the output very carefully. Each type has very subtle differences between them.

You only have 4 tests but there are 6 different styles. Also I can tell that at least one of your tests is outputting incorrect values.

webschik:
@Dan503 , but there were 3 cases for all values without specified `direction` (so I treat them as `direction: ltr`), I've added new cases for `horizontal-tb`, `vertical-rl`, `vertical-lr` with `direction: rtl`

Dan503:
Ok i did a bit of my own testing, the prefixes do seem to be getting applied even if IE11 is the only browser that needs supporting so that's good.

There is still a need to fix your tests though before this can be merged.

Dan503:
In the [original issue](https://github.com/postcss/autoprefixer/issues/1152), look under the "Examples of every combination" heading. There are 6 different styles there that all need to be tested for. They are all slightly different from one another.

webschik:
"Modern" `writing-mode` has 3 values that relate on `direction ltr/rtl` in IE:
```css
writing-mode: horizontal-tb;
writing-mode: vertical-rl;
writing-mode: vertical-lr;
```

So, there are 6 possible cases (3 for each direction):

```css
.one {
    writing-mode: horizontal-tb;
    /* optional direction: ltr; */
}

.two {
    writing-mode: horizontal-tb;
    direction: rtl;
}

.three {
    writing-mode: vertical-rl;
    /* optional direction: ltr; */
}

.four {
    writing-mode: vertical-rl;
    direction: rtl;
}

.five {
    writing-mode: vertical-lr;
    /* optional direction: ltr; */
}

.six {
    writing-mode: vertical-lr;
    direction: rtl;
}
```

we have all of them in [test/cases/writing-mode.css](https://github.com/webschik/autoprefixer/blob/ms-writing-mode/test/cases/writing-mode.css)

Dan503:
I'm actually at the coast. I was going to come home tonight but I'm staying an extra night.

ai:
Released in 9.4.4

PR #1182, opened 2019-01-20
Fix 'implement' typo

Aljullu:
I think there was a small typo: `impliment` instead of `implement`.


ai:
Thanks. Released in 9.4.6.

Issue #1199, opened 2019-03-05
Add support for `mask-composite`


iamvdo:
Reading this article this morning and wondering if this could be added to autoprefixer:

https://css-tricks.com/mask-compositing-the-crash-course/

Especially interesting here is the “Extending support” part

First
```css
mask-composite: add | substract | intersect | exclude
```
should add
```css
-webkit-mask-composite: source-over | source-out | source-in | xor
```
Then, cannot use old keywords in shorthand, so
```css
mask: url(...) add, url(...)
```
should add
```css
-webkit-mask: url(...), url(...);
-webkit-mask-composite: source-over;
```
Finally, if keyword is `source-in`, add `, xor` (but only if one keyword?), so
```css
mask: url(...) intersect, url(...)
```
should add
```css
-webkit-mask: url(...), url(...);
-webkit-mask-composite: source-in, xor;
```

ai:
Seems like we can use https://caniuse.com/#feat=css-masks

ai:
Fixed by https://github.com/postcss/autoprefixer/pull/1202

ai:
Released in 9.5

PR #1202, opened 2019-03-13
Add mask-composite

semeleven:
Fix for #1199
Some moments
1. if keyword is ```source-in```, add ```, xor``` ones. Even has many ```source-in```. Here [example](https://codepen.io/anon/pen/MxOvwj)
2. In the output I have warn from VS Code about lost mask-composite without prefix. What do you think? I think, don't need touch original mask value


ai:
😍 Do you have Twitter and VK to promote you in Autoprefixer accounts?

semeleven:
@ai thank you) My Twitter - https://twitter.com/sem_eleven, VK - https://vk.com/sem_eleven

ai:
Released in 9.5

Issue #1203, opened 2019-03-14
Test fail in OS Windows


Gibr:
When test run in OS Windows it causes fail:
![](http://dl4.joxi.net/drive/2019/03/14/0021/0002/1396738/38/b80a04cdc4.jpg)
Test indicates problem with '}' and as i understand it occurs by difference with chars 13 + 10 (carriage return + line feed) in `output` and just char 10 in `result.css`.
This problem shows only for next tests:
- `prefixes @viewport`
- `has actual example in docs`
- `prevents doubling prefixes`
- `hacks › supports all fullscreens`

ai:
Feel free to send PR

PR #1204, opened 2019-03-14
Fix tests for OS Windows

Gibr:
According to issue #1203 
Excludes '\r' symbol, tests compare uniform values of `output` and `result` variables.


ai:
Can you define this function once in the beginning of the test.

Gibr:
Function moved to the biginning of the test.

ai:
Great. Thanks.

Here is also the way to make it faster 3b857c2

PR #1218, opened 2019-05-28
Removed dead link 'No prefixes in production'

idoo:
I've found that readme contains a dead link that probably should be removed


ai:
Thanks 👍

PR #1253, opened 2019-10-08
Bugfix/separate selectors

fanich37:
Not sure about implementation at all but that's how I see it from my understanding of how the PostCSS and Autoprefixer works :relaxed:

### DIFF
--- a/lib/selector.js
+++ b/lib/selector.js
@@ -3,10 +3,16 @@ let Prefixer = require('./prefixer')
 let Browsers = require('./browsers')
 let utils = require('./utils')
 
+const SELECTORS_CACHE = {}
+
 class Selector extends Prefixer {
   constructor (name, prefixes, all) {
     super(name, prefixes, all)
     this.regexpCache = {}
+    SELECTORS_CACHE[name] = {
+      possible: this.possible(),
+      replace: this.replace.bind(this)
+    }
   }
 
   /**
@@ -57,8 +63,29 @@ class Selector extends Prefixer {
     }
 
     let prefixeds = {}
-    for (let prefix of this.possible()) {
-      prefixeds[prefix] = this.replace(rule.selector, prefix)
+
+    if (rule.selector.includes(',')) {
+      let toProcess = rule.selector
+        .split(',')
+        .filter(el => el.includes(':'))
+        .map(el => el.trim())
+
+      Object.keys(SELECTORS_CACHE).forEach(name => {
+        let selector = toProcess.find(el => el.includes(name))
+        if (selector) {
+          prefixeds[name] = {}
+          SELECTORS_CACHE[name].possible.forEach(prefix => {
+            prefixeds[name][prefix] = SELECTORS_CACHE[name].replace(
+              selector, prefix)
+          })
+        }
+      })
+    } else {
+      prefixeds[this.name] = {}
+
+      for (let prefix of this.possible()) {
+        prefixeds[this.name][prefix] = this.replace(rule.selector, prefix)
+      }
     }
 
     rule._autoprefixerPrefixeds = prefixeds
@@ -79,8 +106,8 @@ class Selector extends Prefixer {
       }
 
       let some = false
-      for (let key in prefixeds) {
-        let prefixed = prefixeds[key]
+      for (let key in prefixeds[this.name]) {
+        let prefixed = prefixeds[this.name][key]
         if (before.selector === prefixed) {
           if (prefix === key) {
             return true
@@ -117,7 +144,7 @@ class Selector extends Prefixer {
       return
     }
 
-    let cloned = this.clone(rule, { selector: prefixeds[prefix] })
+    let cloned = this.clone(rule, { selector: prefixeds[this.name][prefix] })
     rule.parent.insertBefore(rule, cloned)
   }
 
--- a/lib/selector.js
+++ b/lib/selector.js
@@ -1,4 +1,4 @@
-let list = require('postcss').list
+let { list } = require('postcss')
 
 let OldSelector = require('./old-selector')
 let Prefixer = require('./prefixer')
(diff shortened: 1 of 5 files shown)


ai:
Wow, great. This fix is great (and looks like correct) but should have a double check (and I am too drunk right now). I will try to check it this week.

lib/selector.js
```diff
@@ -3,10 +3,16 @@ let Prefixer = require('./prefixer')
 let Browsers = require('./browsers')
 let utils = require('./utils')
 
+const SELECTORS_CACHE = {}
+
 class Selector extends Prefixer {
   constructor (name, prefixes, all) {
     super(name, prefixes, all)
     this.regexpCache = {}
+    SELECTORS_CACHE[name] = {
```
ai: How we will lean this cache? Will we have a problem if PostCSS will process many very different files (or even different versions of the same files) during the same Node.js session (so the cache will be alive?)

If you want a per-document cache, you can use `node.root().autoprefixerSelectors` cache. It will be attached only for the same CSS root.

fanich37:
> Wow, great. This fix is great (and looks like correct) but should have a double check (and I am too drunk right now). I will try to check it this week.

Thank for your fast reply! It would be awesome if you found some time to test it more precisely. But it's definitely not the fix to rush with :+1:

lib/selector.js
fanich37: Oops, that's a really good question. I need to say I didn't even thought about such cases. I'll try to check them asap. Thanks!

fanich37: I haven't tested it yet. I just thought that ```SELECTORS_CACHE``` is filled with data during the file/files preprocessing. So probably it should work fine with multiple files. But I'm not 100% sure. By the way the naming ```SELECTORS_CACHE``` is kinda misleading. It's not a cache, it's just a collection.

ai: Can we replace it to `this.selectors` since we use only instance methods and data there?

fanich37: Nope, it wouldn't work in such implementation. The idea with this global const is to keep all possible selectors with all relevant prefixes. And with ```this.selectors``` we only get relevant selectors for ```this.name```. But rule might contain several selectors.

I like your propose about using ```node.root().autoprefixerSelectors```. I haven't checked it yet and actually I don't know how it works but may be It could help to get rid of ```SELECTORS_CACHE```.

ai: I am afraid of putting an instance method to the global cache. If this method is not related to instance, we can move it to global function with Memoization cache.

fanich37: >I am afraid of putting an instance method to the global cache.

Yep, I don't like it either.
>If this method is not related to instance, we can move it to global function with Memoization cache.

Could you please explain a bit more about that? Probably I don't get your point completely. Or maybe I just need to sleep a bit more...

In any case thanks for your time and thoughts about this PR!

ai: > Could you please explain a bit more about that?

Memoization is a pattern like this

```js
let cache = { }
function method (arg) {
  return … // Do heavy stuff
}
function methodCached (arg) {
  if (!cache[args]) {
    cache[args] = method(arg)
  }
  return cache[args]
}
```

It means if you have a pure function (which returns the same result on the same argument) you can cache the result in the value.

I thought that this pattern can be useful here, but found that I may understand the algorithm wrong (sorry, have too many things to do recently). Can you explain what do you do in `SELECTORS_CACHE`?

fanich37: Thank you for the explanation. I thought that your idea was to hide ```SELECTORS_CACHE``` in a closure. Glad I asked.

My idea is: the ```SELECTORS_CACHE``` is a list of all possible selectors, all possible related prefixes and related ```replace``` method of the instance. It's filled with data during the file preprocessing on any initialization of ```class Selector``` (and it's descendants). So after preprocessing we get smth like this:
```js
SELECTORS_CACHE = {
    '::placeholder': {
        possible: ['-moz-', '-webkit-'], // all possible prefixes for '::placeholder'
        replace, // consider it as a replace method of new Selector('::placeholder'),
    },
    ':read-only': {
        possible: ['-moz-'], // all possible prefixes for ':read-only'
        replace, // consider it as a replace method of new Selector(':read-only'),
    }
    // and so on
};
```
Why do I need to keep all that data in one place? Cause later when ```Autoprefixer``` adds prefixes there could be rule with several selectors, for example ```.a::placeholder, .b:read-only```, and while the rule is processed with, the ```::placeholder``` instance, we need access not only to ```::placeholder``` prefixes and ```replace``` method but to ```:read-only``` prefixes and ```replace``` method too.

Hope I described my idea clear enough. And probably there is another (more proper) way to get the same behavior but I don't know it 😃

ai: If `replace` doesn’t not depends on instance data, can we move it to pure function? `replaceSelector()`

If possible selectors don’t rely on Browserslist config (because it is all possible prefixes) can we move it to pure function too? `getSelectorPossiblePrefixes()`

Then we can add Memoization to these two functions and call them directly from `lib/proccessor` or `lib/selector`. It could make code easier. Right now it is too complex to understand it without looking deeper and witching between files (I am afrading that next developers will not understand it too).

fanich37: I can try to do so. My concern is that ```replace``` method uses the instance method ```this.prefixed(prefix)``` inside. But hacks ```::placeholder``` and ```:fullscreen``` are overriding it. The ```::placeholder``` also overrides ```this.possible()``` method.

ai: > My concern is that replace method uses the instance method this.prefixed(prefix) inside. 

If it really rely on instance, you can't detouch it from this instance.

Don't forget that PostCSS can be called on many different CSS files.

fanich37: > If it really rely on instance, you can't detouch it from this instance.

That's make sense.
I'll try to implement your idea.

> Don't forget that PostCSS can be called on many different CSS files.

Can it process different files with different browser settings?

ai: > Can it process different files with different browser settings?

Yeap. The same instance can be used for different target browsers or different whitespace settings.

It will be safer just to think that every PostCSS call must be independent from each other.

fanich37: I don't know what was I thinking when I decided to use global ```const SELECTORS_CACHE```. Each instance of class ```Selector``` gets property ```all``` on its init. And ```all``` is just the ```Prefixes``` instance that already contains all the data we need to process grouping rules :smile:

lib/selector.js
```diff
@@ -1,3 +1,5 @@
+let list = require('postcss').list
```
ai:
```suggestion
let { list } = require('postcss')
```

lib/selector.js
```diff
@@ -57,8 +59,35 @@ class Selector extends Prefixer {
...
+      this.all.add.selectors.forEach(selector => {
+        let name = selector.name
+        let selToProc = separatedRule.filter(el => el.includes(name))
+
+        if (selToProc.length) {
+          let possible = selector.possible()
+          prefixeds[name] = {}
+
+          possible.forEach(prefix => {
+            if (selToProc.length > 1) {
```
ai: I think we can remove this `if`. Array with a single element will have the same result after processing with `map` and `join`. It will be a little slower, but it is not a problem here since the main idea is to reduce code complexity.

lib/selector.js
```diff
@@ -57,8 +59,35 @@ class Selector extends Prefixer {
     }
 
     let prefixeds = {}
-    for (let prefix of this.possible()) {
-      prefixeds[prefix] = this.replace(rule.selector, prefix)
+
+    if (rule.selector.includes(',')) {
```
ai: I completely lost the idea behind this method (but I wrote it :sob:).

Can you describe (you can do it in Russian) what do we do here?

fanich37: I'll try to do it in english :crossed_fingers: 

This method collects all possible prefixes for the selector, creates prefixed version of the ```rule.selector```, caches the result and adds it to the rule. So the rule looks like this at the end (copied from master branch):
```js
Rule {
    type: 'rule',
    selector: '.a, .a .b::placeholder, a:read-only',
    // lots of properties
    _autoprefixerPrefixeds: {
        '-webkit-': '.a, .a .b::-webkit-input-placeholder, a:read-only',
        '-moz-': '.a, .a .b::-moz-placeholder, a:read-only',
        '-ms-': '.a, .a .b::-ms-input-placeholder, a:read-only',
        '-o-': '.a, .a .b::-o-placeholder, a:read-only',
        '-moz- old': '.a, .a .b:-moz-placeholder, a:read-only',
        '-ms- old': '.a, .a .b:-ms-input-placeholder, a:read-only' }
    }
}
```
As you can see ```:read-only``` is left without prefixes.

lib/selector.js
fanich37: Replaced with your suggestion :+1:

lib/selector.js
fanich37: Removed it.

lib/selector.js
fanich37: I refactored code, and I think now it's much more easier to reason about. Plus, I decided that each selector has to process only that part of rule that contains the selector and not the whole rule as I did before.

ai:
Great! Now it became much simpler.

ai:
Thanks. Released in 9.6.5. Great work in cleaning PR.

fanich37:
> Thanks. Released in 9.6.5. Great work in cleaning PR.

Thanks for your review and questions that helped me to find the proper solutions 👍

Zevran:
I have now an error with pseudo element `placeholder` since this path. See https://github.com/cssnano/cssnano/issues/828

PR #1260, opened 2019-10-26
Feature/vendor specific transition

### LINKED ISSUE #1217: Autoprefixer generically adds prefixes to browser-specific pseudo-classes
Hi, how to block autoprefixer from adding -webkit prefixes to a Mozilla specific extension ?

Here are some examples to make my point :

```css
.custom-range::-moz-range-thumb {
    -webkit-transition: none;
    transition: none
}
```

`-webkit-transition` has nothing to do here as it is a Mozilla specific selector (`::-moz-range-thumb`).

Another crazy output happens with keyframes and transitions, as Autoprefixer creates useless variants

```css
  -webkit-transition: -webkit-box-shadow 0.1s ease-out;
  transition: -webkit-box-shadow 0.1s ease-out;
  transition: box-shadow 0.1s ease-out;
  transition: box-shadow 0.1s ease-out, -webkit-box-shadow 0.1s ease-out;
```

Instead of just

```css
  -webkit-transition: -webkit-box-shadow 0.1s ease-out;
  transition: box-shadow 0.1s ease-out;
```

How to fix this ? Thanks for your help. And thanks a lot for Autoprefixer !!!

fanich37:
This PR should close #1217 . The issue is about ```transition``` property was prefixed with all of the prefixes even though not all of them are required (like ```-moz-``` prefixes in ```-webkit-``` rule).

### DIFF
--- a/lib/transition.js
+++ b/lib/transition.js
@@ -314,17 +314,15 @@ class Transition {
      */
     inVendorSpec (decl) {
       let { parent } = decl
-      let { selector, type } = parent
 
-      if (decl._autoprefixerInVendorSpecific !== undefined) {
-        return decl._autoprefixerInVendorSpecific
+      if (parent.type !== 'rule') {
+        return false
+      } else if (!parent.selector.includes(':-')) {
+        return false
+      } else {
+        return !this.prefixes.add.selectors
+          .find(s => parent.selector.includes(s.name))
       }
-
-      let inSpecific = type === 'rule' && selector.includes(':-') &&
-        !this.prefixes.add.selectors.find(s => selector.includes(s.name))
-      decl._autoprefixerInVendorSpecific = inSpecific
-
-      return inSpecific
     }
 }
 
--- a/lib/transition.js
+++ b/lib/transition.js
@@ -9,13 +9,24 @@ class Transition {
       this.prefixes = prefixes
     }
 
+    process (decl, result) {
+      let add = this.prefixes.add[decl.prop]
+      let declPrefixes = (add && add.prefixes) || []
+      let inSpecific = this.inVendorSpec(decl)
+
+      if (inSpecific) {
+        let prefix = decl.parent.selector.match(/:(-\w+-)/)[1]
+        declPrefixes = declPrefixes.filter(p => p === prefix)
+      }
+
+      this.add(decl, result, declPrefixes)
+    }
+
     /**
      * Process transition and add prefixes for all necessary properties
      */
-    add (decl, result) {
+    add (decl, result, declPrefixes) {
       let prefix, prop
-      let add = this.prefixes.add[decl.prop]
-      let declPrefixes = (add && add.prefixes) || []
 
       let params = this.parse(decl.value)
       let names = params.map(i => this.findProp(i))
@@ -293,6 +304,24 @@ class Transition {
         }
       }
       return undefined
+    }
+
+    /**
+     * Check if transition prop is inside vendor specific rule
+     */
+    inVendorSpec (decl) {
+      let { parent } = decl
+      let { selector, type } = parent
+
+      if (decl._autoprefixerSpecific !== undefined) {
+        return decl._autoprefixerSpecific
+      }
+
+      let inSpecific = type === 'rule' && selector.includes(':-') &&
+        !this.prefixes.add.selectors.find(s => selector.includes(s.name))
+      decl._autoprefixerSpecific = inSpecific
+
+      return inSpecific
     }
 }
 
--- a/lib/transition.js
+++ b/lib/transition.js
@@ -36,7 +36,9 @@ class Transition {
         if (!prefixer || !prefixer.prefixes) continue
 
         for (prefix of prefixer.prefixes) {
-          if (vendorPrefixes && !vendorPrefixes.includes(prefix)) continue
+          if (vendorPrefixes && !vendorPrefixes.some(p => prefix.includes(p))) {
+            continue
+          }
 
           let prefixed = this.prefixes.prefixed(prop, prefix)
           if (prefixed !== '-ms-transform' && !names.includes(prefixed)) {
--- a/lib/transition.js
+++ b/lib/transition.js
@@ -41,9 +41,12 @@ class Transition {
         if (prop[0] === '-') continue
 
         let prefixer = this.prefixes.add[prop]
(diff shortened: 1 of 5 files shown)


lib/transition.js
```diff
@@ -294,6 +308,24 @@ class Transition {
...
     }
+
+    /**
+     * Check if transition prop is inside vendor specific rule
+     */
+    inVendorSpec (decl) {
+      let { parent } = decl
+      let { selector, type } = parent
+
+      if (decl._autoprefixerInVendorSpecific !== undefined) {
```
ai: This cache is not really necessary, since it content is pretty simple to calculate

lib/transition.js
```diff
@@ -294,6 +308,24 @@ class Transition {
       }
       return undefined
     }
+
+    /**
+     * Check if transition prop is inside vendor specific rule
+     */
+    inVendorSpec (decl) {
+      let { parent } = decl
+      let { selector, type } = parent
```
ai: I prefer to have explicit `parent.type`. It is easy to misunderstand code above by thinking that `type` is `decl.type`.

lib/transition.js
```diff
@@ -294,6 +308,24 @@ class Transition {
...
+
+      if (decl._autoprefixerInVendorSpecific !== undefined) {
+        return decl._autoprefixerInVendorSpecific
+      }
+
+      let inSpecific = type === 'rule' && selector.includes(':-') &&
```
ai: I think this form will be more readable:

```js
if (parent.type !== 'rule') {
  return false
} else if (!parent.selector.includes(':-')) {
  return false
} else {
  return this.prefixes.add.selectors.find(s => parent.selector.includes(s.name))
}
```

lib/transition.js
fanich37: Removed it.

lib/transition.js
fanich37: Agree with you, it could be misleading. Rewrote with ```parent.type``` and ```parent.selector```.

lib/transition.js
fanich37: Nice approach! Rewrote it.

lib/transition.js
```diff
@@ -294,6 +308,22 @@ class Transition {
...
+    inVendorSpec (decl) {
+      let { parent } = decl
+
+      if (parent.type !== 'rule') {
+        return false
+      } else if (!parent.selector.includes(':-')) {
+        return false
+      } else {
+        return !this.prefixes.add.selectors
+          .find(s => parent.selector.includes(s.name))
```
ai: 1. `.some()` will be faster (and more readable)
2. Why do we need to test prefixes from Autoprefixer. Some prefixes are non-standard and Autoprefixer doesn’t support it (for instance, `::-webkit-scrollbar`).
3. We can return `declPrefixes` from `inVendorSpec` to make code more simple.

lib/transition.js
```diff
@@ -294,6 +308,22 @@ class Transition {
       }
       return undefined
     }
+
+    /**
+     * Check if transition prop is inside vendor specific rule
+     */
+    inVendorSpec (decl) {
```
ai: `*Spec` reminds me test’s spec. I think `ruleVendorPrefixes()` will be better name (and we return `declPrefixes` here)

lib/transition.js
```diff
@@ -30,9 +41,12 @@ class Transition {
         if (prop[0] === '-') continue
 
         let prefixer = this.prefixes.add[prop]
+        let inSpecific = this.inVendorSpec(decl)
```
ai: If we will merge `declPrefixes` and `inSpecific` generation we will be able to avoid the double call of `inVendorSpec`

lib/transition.js
fanich37: > Why do we need to test prefixes from Autoprefixer.

Cause we don't need to handle standard selectors like ```:read-only``` and so on. The decls inside this selectors have to prefixed with all relevant prefixes.

> Some prefixes are non-standard and Autoprefixer doesn’t support it (for instance, ::-webkit-scrollbar).

And decls inside this selector will be prefixed only with ```-webkit-``` prefix.

ai: > Cause we don't need to handle standard selectors like :read-only and so on. The decls inside this selectors have to prefixed with all relevant prefixes.

You can use `Browsers.prefixes()` to get all prefixes and then check them in selector:

```js
} else {
  let prefixes = Browsers.prefixes().filter(i => parent.selector.includes(':-' + i))
  return prefixes.length > 0 ? prefixes : false
}
```

lib/transition.js
fanich37: I completely remove ```process``` method cause now it seems to be useless at all.

lib/transition.js
fanich37: Yes, it's more clear. Renamed it.

lib/transition.js
fanich37: Completely forgot about ```Browsers```. Thanks!

lib/transition.js
```diff
@@ -33,6 +36,8 @@ class Transition {
         if (!prefixer || !prefixer.prefixes) continue
 
         for (prefix of prefixer.prefixes) {
+          if (vendorPrefixes && !vendorPrefixes.includes(prefix)) continue
```
ai: We can always have array in `vendorPrefixes` (by returning `[]` instead of `false`) and about this extra check.

fanich37: I like the idea of returning ```false``` cause here we need to check whether rule is vendor specific. If it's not - we don't have to check the prefix. By the way I just found out the check itself 
```js
!vendorPrefixes.includes(prefix)
```
works incorrect. Prefix might be ```-webkit- 2009``` for example. I'll fix it.

lib/transition.js
```diff
@@ -15,7 +17,8 @@ class Transition {
     add (decl, result) {
       let prefix, prop
       let add = this.prefixes.add[decl.prop]
-      let declPrefixes = (add && add.prefixes) || []
+      let vendorPrefixes = this.ruleVendorPrefixes(decl)
+      let declPrefixes = vendorPrefixes || (add && add.prefixes) || []
```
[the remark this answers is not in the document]
fanich37: I described it [here](https://github.com/postcss/autoprefixer/pull/1260#discussion_r340445851).

ai:
Released in 9.7.1

PR #1272, opened 2019-12-08
Add test for at-rules

isolovev:


test/autoprefixer.test.js
```diff
@@ -547,6 +547,13 @@ it('works with CSS Modules', () => {
   postcss([autoprefixer()]).process(':export { selectors: _1q6ho_2 }').css
 })
 
+it('ignore prefix in vendor at rules', () => {
+  let input = read('at-rules')
+  let output = read('at-rules.out')
+  let result = postcss([autoprefixer({ cascade: false })]).process(input)
```
ai: `autoprefixer` will use current browsers and could change output in the future

test/autoprefixer.test.js
```diff
@@ -547,6 +547,13 @@ it('works with CSS Modules', () => {
   postcss([autoprefixer()]).process(':export { selectors: _1q6ho_2 }').css
 })
 
+it('ignore prefix in vendor at rules', () => {
+  let input = read('at-rules')
```
ai: We have a shortcut to do this test. You can find it in `hacks` secltion in this test file.

test/cases/at-rules.out.css
```diff
@@ -0,0 +1,33 @@
+body {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
```
ai: I recommend enabling `cascade` here to make test readable and more standard

isolovev: I fixed that.

test/autoprefixer.test.js
isolovev: I fixed that.

test/autoprefixer.test.js
isolovev: I fixed that.

PR #1277, opened 2020-01-21
Fix link to PostCSS warning API docs

barbeque:
The PostCSS documentation has moved, and this link moved with it. I believe this is the new page for the `Warning` class. Thanks for the great library!


ai:
Thanks for finding it and fixing :+1:

PR #1297, opened 2020-03-27
Added another Text Editor (VSCode)

kanlukasz:


ai:
What is a different between these 2 plugins?

PR #1326, opened 2020-06-19
chore: update "example" expectant

lukeed:
Browserslist was recently updated (9c4c7d8b40f57f388cfa996e9d0f5a55663fbc93) which must have updated the default list of resolved browsers.

### DIFF
--- a/test/cases/example.out.css
+++ b/test/cases/example.out.css
@@ -1,6 +1,3 @@
-::-webkit-input-placeholder {
-  color: gray;
-}
 ::-moz-placeholder {
   color: gray;
 }


ai:
You also need to fix `README.md` (I use this test to find main example changes).

---

If you are trying to move `chalk` to `kleur` here, note that Autoprefixer’s `master` is used for 9.x branch releases (which must support Node.js 6).

We didn’t open a new branch for 10.x (I will do it only when PostCSS 8.0 work will be finished).

lukeed:
Ah, I just opened that PR. I can use `kleur@3` which supports Node 6.x, or I can patch kleur@4 to lower engines requirement. The code will still run correctly in Node 6

lukeed:
This passes in CI, so there's no reason to change it here. Must be something fishy on my machine. 🤷‍♂️

PR #1334, opened 2020-07-11
Fix for #1333 - No warnings when grid is turned off; Also added a subgrid warning

### LINKED ISSUE #1333: Autoprefixer processing comment to locally turn off grid polyfill continues to report IE related warnings
Given the following snippit of CSS

```css
.usps > ul {
  display : flex; 
  // ...
}

@supports(grid-template-columns: repeat(auto-fit, minmax(200px, max-content))) {
  .usps > ul {
    /* autoprefixer grid: off */
    display               : grid;
    grid-gap              : 20px 10px;
    grid-template-columns : repeat(auto-fit, minmax(200px, max-content));
    justify-content       : center;
  }
}
```

Even though it is using the `/* autoprefixer grid: off */` processing comment, this particular block continues to raise a warning that " `IE does not support justify-content on grid containers` ." when conditionally applying grid styling.

Elevating the processing comment to the scope of the `@supports` block itself yields the same result. It continues to report warnings related to IE's grid implementation. Only using `/* autoprefixer: off */` in these places works as expected.

While that suffices in the current situation for my problem, it is highly non-desired that the entirety of Autoprefixer has to be turned off. This would start causing problems once these rules end up containing properties which _do_ need prefixing, requiring that e.g. an artificially split CSS rule be maintained where autoprefixer can still be turned off.

**Autoprefixer version** : 9.8.4

Dan503:
- Fixes #1333 - It no longer triggers grid warnings if `/* autoprefixer grid: off */` is used
- It will now warn that subgrid is not supported in IE if someone tries to use `subgrid` as a value in a declaration
- I added "npx" to the front of the "gulp play" command in the playground css comment to prevent people getting confused if they don't have gulp installed globally.

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -146,7 +146,7 @@ class Processor {
           { node: decl }
         )
       } else {
-        if (gridPrefixes) {
+        if (gridPrefixes && this.gridStatus(decl, result)) {
           if (/^(align|justify|place)-items$/.test(prop) && insideGrid(decl)) {
             let fixed = prop.replace('-items', '-self')
             result.warn(
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -147,6 +147,12 @@ class Processor {
         )
       } else {
         if (gridPrefixes && this.gridStatus(decl, result)) {
+          if (decl.value === 'subgrid') {
+            result.warn(
+              'IE does not support subgrid',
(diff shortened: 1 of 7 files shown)


ai:
Thanks!

ai:
Released in 9.8.5.

PR #1363, opened 2020-09-26
Simplify uniq util

SukkaW:
Simplify `uniq(array)` util with `Set` & spread syntax. All tests are passed.

`Set` is supported since Node.js 0.13 while spread syntax is fully supported since Node.js 8.3.0.


ai:
Thanks

PR #1365, opened 2020-09-26
Refactor: introduce a cache util

SukkaW:
The PR brings up the `Cache` util to reuse the code.

All tests are passed. The coverage of `util.js` remains 100%.

### DIFF
--- a/lib/autoprefixer.js
+++ b/lib/autoprefixer.js
@@ -6,6 +6,7 @@ let Browsers = require('./browsers')
 let Prefixes = require('./prefixes')
 let data = require('../data/prefixes')
 let info = require('./info')
+let { Cache } = require('./utils')
 
 const WARNING =
   '\n' +
@@ -26,7 +27,7 @@ function isPlainObject (obj) {
   return Object.prototype.toString.apply(obj) === '[object Object]'
 }
 
-let cache = new Map()
+let cache = new Cache()
 
 function timeCapsule (result, prefixes) {
   if (prefixes.browsers.selected.length === 0) {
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -1,5 +1,25 @@
 let { list } = require('postcss')
 
+class Cache {
+  constructor () {
+    this.map = new Map()
```
SukkaW, line 5: `map` should be a "private" field as it shouldn't be exposed, however `private` is only supported since Node.js 12 so I just leave it here.

lib/utils.js
```diff
@@ -1,5 +1,26 @@
 let { list } = require('postcss')
 
+class Cache {
+  constructor () {
+    // FIXME: private class field is supported for Node.js 12 and newer
+    this.map = new Map()
```
SukkaW, line 6: `map` should be a "private" field as it shouldn't be exposed, however `private` is only supported since Node.js 12. A `FIXME` message has been added.

lib/autoprefixer.js
```diff
@@ -6,6 +6,7 @@ let Browsers = require('./browsers')
 let Prefixes = require('./prefixes')
 let data = require('../data/prefixes')
 let info = require('./info')
+let { Cache } = require('./utils')
```
ai, line 9: Can you put it to the first line to order imports by name length `{ Cache } > info`

lib/utils.js
```diff
@@ -1,5 +1,26 @@
 let { list } = require('postcss')
 
+class Cache {
+  constructor () {
```
ai, line 4: What does this wrap give to us? For me, it will just slow down the calculation.

I think for the few use cases, isolation from the wrapper class is not necessary.

PR #1382, opened 2021-01-03
Add typings

Semigradsky:
Add typings based on typings from DefinitelyTyped.

### DIFF
--- a/lib/autoprefixer.d.ts
+++ b/lib/autoprefixer.d.ts
@@ -1,0 +1,95 @@
+import { Plugin } from 'postcss'
+import { Stats } from 'browserslist'
+
+declare function autoprefixer<T extends string[]> (
+  ...args: [...T, autoprefixer.Options]
+): Plugin & autoprefixer.ExportedAPI
+
+declare function autoprefixer (
+  browsers: string[],
+  options?: autoprefixer.Options
+): Plugin & autoprefixer.ExportedAPI
+
+declare function autoprefixer (
+  options?: autoprefixer.Options
+): Plugin & autoprefixer.ExportedAPI
+
+declare namespace autoprefixer {
+  type GridValue = 'autoplace' | 'no-autoplace'
+
+  interface Options {
+    /** environment for `Browserslist` */
+    env?: string
+
+    /** should Autoprefixer use Visual Cascade, if CSS is uncompressed */
+    cascade?: boolean
+
+    /** should Autoprefixer add prefixes. */
+    add?: boolean
+
+    /** should Autoprefixer [remove outdated] prefixes */
+    remove?: boolean
+
+    /** should Autoprefixer add prefixes for @supports parameters. */
+    supports?: boolean
+
+    /** should Autoprefixer add prefixes for flexbox properties */
+    flexbox?: boolean | 'no-2009'
+
+    /** should Autoprefixer add IE 10-11 prefixes for Grid Layout properties */
+    grid?: boolean | GridValue
+
+    /** custom usage statistics for > 10% in my stats browsers query */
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -74,13 +81,15 @@
...
+    "extends": "@logux/eslint-config/ts",
     "rules": {
       "security/detect-non-literal-regexp": "off",
       "security/detect-unsafe-regex": "off",
       "node/no-missing-require": "off",
       "no-unused-expressions": "off",
-      "no-unused-vars": "off"
+      "no-unused-vars": "off",
+      "no-redeclare": "off",
+      "@typescript-eslint/no-redeclare": ["error"]
```
ai, line 92:
```suggestion
      "@typescript-eslint/no-redeclare": "error"
```

ai: What error it prevents in Autoprefixer?

lib/autoprefixer.d.ts
```diff
@@ -0,0 +1,107 @@
...
+  /** Inspect with default Autoprefixer */
+  let info: ExportedAPI['info']
+
+  let postcss: true
+}
+
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv {
+      AUTOPREFIXER_GRID?: autoprefixer.GridValue
```
ai, line 102: It can be also `undefined`

lib/autoprefixer.d.ts
```diff
@@ -0,0 +1,107 @@
+import { Plugin } from 'postcss'
+import { Stats } from 'browserslist'
+
+declare function autoprefixer<T extends string[]> (
```
ai, line 4: Why do we need `T`?

lib/autoprefixer.d.ts
```diff
@@ -0,0 +1,107 @@
...
+    stats?: Stats
+
+    /** @deprecated 'Change `browser` option to `overrideBrowserslist` in Autoprefixer */
+    browser?: string[] | string
+
+    /**
+     * @deprecated Replace Autoprefixer `browsers` option to Browserslist config.
+     * Use `browserslist` key in `package.json` or `.browserslistrc` file.
+     */
+    browsers?: string[] | string
```
ai, line 52: I prefer to remove deprecated options from types

package.json
Semigradsky: `no-redeclare` rule forbid to make function overloading. Maybe it is better to move rules to `@logux/eslint-config`:
```
      "no-redeclare": "off",
      "@typescript-eslint/no-redeclare": "error"
```

lib/autoprefixer.d.ts
Semigradsky: It is okay, `AUTOPREFIXER_GRID?` can be `undefined`.

package.json
ai: Where it could happen?

lib/autoprefixer.d.ts
Semigradsky: That is the way to implement leading rest parameter in TypeScript 4. In future it will be looks much better: https://github.com/microsoft/TypeScript/pull/41544

package.json
Semigradsky: In any place when you will have function overloading or interface declaration merging. `no-redeclare` doesn't understand that it is not "redeclare", `@typescript-eslint/no-redeclare` handles these cases.

ai: Thanks for the suggestion. I released ESLint config 44.0.1 and updated config in Autoprefixer’s `master`.

Can you revert config changes (and we are ready for the release)?

ai:
Released in 10.2

PR #1383, opened 2021-01-05
fix CONTRIBUTING.md formatting

Sheraff:
To make it easier for potential new contributors, I propose the following changes:
- fix logo src path: seems like a simple typo from copy-pasting from README.md that is in the parent folder, so I just added `../`
- code blocks indentation: seems like github flavored markdown needs a different indentation from the one that was used, i just used syntax highlighting to make sure github could parse the code blocks correctly.


ai:
Thanks!

Issue #1384, opened 2021-01-06
Minimize # of "Replace transition-property to transition" warnings


Sheraff:
This issue is aimed at increasing the accuracy of the following warning in order to avoid "error numbness":

> Replace transition-property to transition, because Autoprefixer could not support any cases of transition-property and other transition-*

I see in the [checkForWarning function in transition.js](https://github.com/postcss/autoprefixer/blob/main/lib/transition.js#L116) that in order to emit the warning, we need to have a `transition-property` prop *and* another `transition-*` prop *in the same rule*. Is there a reason why it is so? One could reasonably imagine that, for example, the `transition-delay` is set by another rule. In which case `checkForWarning` would omit the issue.

1. I would argue that in order not to miss any potential issue, the warning should be raised as soon as a `transition-property` prop *with comma separated value list* is found. And the warning should be along the lines of:

   > Replace transition-property with multiple values to transition, because Autoprefixer cannot know which other transition-* properties it is associated with.

   In this case we wouldn't miss any warnings but would emit much more than before, which is why I propose this second change.

2. In order to avoid emitting unnecessary warnings (which developers might end up ignoring) wouldn't it be preferable to only emit the warning *if at least one of the values in the transition-property value list needs to be prefixed*? This would need to take into account the browserslist config, and with the more modern browser scopes we're increasingly using, the number of warnings would go down significantly.

   For example the following source would never raise a warning, whereas right now it does:
   ```scss
   .foo {
     transition-property: color, background-color; // <-- no prefixing needed, no warning
     transition-duration: 1s, 2s;
   }
   ```

   And the following source would only raise a warning *depending on the scope set by browserslist*, whereas now it always does:
   ```scss
   .foo {
     transition-property: color, transform; // <-- prefixing needed sometimes, only warn when necessary
     transition-duration: 1s, 2s;
   }
   ```

   Finally, the following source would now raise a warning (again, depending on browserslist), whereas now it doesn't:
   ```scss
   .foo {
     transition-property: color, transform; // <-- should still warn sometimes
   }
   .bar {
      transition-duration: 1s, 2s; // <-- because this might exist somewhere else
   }
   ```

If you agree with the modifications I propose, I'd be happy to try and submit a pull request, though for the 2nd one I would need to be pointed in the right direction for checking that "at least one of the values in the transition-property value list needs to be prefixed".

ai:
[...]

PR #1385, opened 2021-01-06
check transition-property value list before warn

Sheraff:
This PR answers the issue #1384 (and following discussion). It includes related tests and code comments.

### DIFF
--- a/lib/transition.js
+++ b/lib/transition.js
@@ -118,6 +118,9 @@ class Transition {
       return
     }
 
+    let isPrefixed = false
+    let hasAssociatedProp = false
+
     decl.parent.each(i => {
       if (i.type !== 'decl') {
         return undefined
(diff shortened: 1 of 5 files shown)


test/autoprefixer.test.js
```diff
@@ -361,6 +361,15 @@ it('prefixes transition', () => {
   ])
 })
 
+it('does not raise unnecessary warnings when prefixing transition', () => {
```
ai, line 364: You can use `check()` function to make a test smaller

lib/transition.js
```diff
@@ -118,28 +118,43 @@ class Transition {
...
-      if (list.comma(i.value).length > 1) {
-        decl.warn(
-          result,
-          'Replace transition-property to transition, ' +
-            'because Autoprefixer could not support ' +
-            'any cases of transition-property ' +
-            'and other transition-*'
-        )
+      // check if another transition-* prop in current Rule has comma separated value list
+      if (values.length > 1) {
```
ai, line 143: Can you replace it to simpler

```js
hasAssociatedProp = values.length > 1
```

Sheraff: Not really, because a subsequent transition-* property (without a comma separated value list) might override it in a following iteration.

I could replace it with 
```js
hasAssociatedProp = hasAssociatedProp || values.length > 1
```

ai:
Tests are falling now

Sheraff:
The requested changes are done, and I merged the newer commits (new typings were "conflicting" without issue).

I also noticed that the autofix linter isn't set up to work on .ts files (which there is now since the new typings merge)
```json
"lint-staged": {
  "*.js": "eslint --fix"
},
```

ai:
Thanks. I added `.ts` to ESLint auto-fix 5bfb37e

I will release these changes next day.

Issue #1391, opened 2021-01-27
Support for "@supports selector"


dvoytenko:
The `@supports selector() {}` form is now supported in Chromium and Mozilla-based browsers. It'd be great if it was supported in PostCSS as well.

https://developer.mozilla.org/en-US/docs/Web/CSS/@supports

ai:
What is the error message?

dvoytenko:
"Unknown node type in node creation"

ai:
PR is welcome. I am busy on another open-source project right now.

cpmsmith:
It looks like this is specific to selectors starting with `:`, i.e. pseudo-classes and pseudo-elements.

```css
/* works: */
@supports selector(a) {}
/* throws error: */
@supports selector(:a) {}
```

ai:
The fix by @Semigradsky was released in 10.2.5

dvoytenko:
Thanks so much for the fix, @Semigradsky !

itf2012:
I'm currently experiencing the issue mentioned by @KaelWD but i'm not seeing/understanding what the resolution is. Can anyone provide any guidance? Thanks.

PR #1398, opened 2021-03-04
Fix error throwing when `supports-selector-fn` has pseudo-classes or pseudo-elements as argument.

Semigradsky:
Related to #1391

### DIFF
--- a/test/supports.test.js
+++ b/test/supports.test.js
@@ -155,4 +155,18 @@ describe('process()', () => {
     supports.process(rule)
     expect(rule.params).toEqual('(color black) and not ((-moz-a: 1) or (a: 1))')
   })
+
+  it("shouldn't throw errors", () => {
+    let rule = { params: 'not selector(:is(a, b))' }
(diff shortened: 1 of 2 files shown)


Semigradsky:
@dvoytenko @cpmsmith Could you please try this change?

test/supports.test.js
```diff
@@ -155,4 +155,12 @@ describe('process()', () => {
     supports.process(rule)
     expect(rule.params).toEqual('(color black) and not ((-moz-a: 1) or (a: 1))')
   })
+
+  it("shouldn't throw errors", () => {
+    let rule = { params: ' (selector( :nth-child(1n of a, b) )) or (c: b(1)) ' }
```
ai, line 160: Can you add also a case without white-space between `selector(` and `:`?

Semigradsky: I have added one more test.

ai:
Released in 10.2.5

PR #1407, opened 2021-05-19
Add console syntax highlighting

Yash-Singh1:


README.md
```diff
@@ -1062,7 +1062,7 @@ If you wish to change the size of a `grid-gap`, you will need to redeclare the g
 Run `npx autoprefixer --info` in your project directory to check
 which browsers are selected and which properties will be prefixed:
 
-```
+```console
```
ai, line 1065:
```suggestion
```sh
```

Yash-Singh1: The `sh` language specifier is made for shell scripts while the `console` one detects the `$` sign and considers that as the prompt and the rest as the output. The `console` one is what is used in this text.

ai: Got it. Thanks. Didn’t know about this difference.

Issue #1318, opened 2020-06-05
Remove Time-traveling warning


malixsys:
Is this necessary? In `dev` there may well be no targets and the warnings will needlessly show up...
https://github.com/postcss/autoprefixer/blob/0283e319ff7bcde0454c959036e0d91ae0bb326f/lib/autoprefixer.js#L43

Timer:
Furthermore, just because a build doesn't require prefixing in the present doesn't mean that a few weeks in the future wouldn't introduce new properties that need prefixing. It'd be a shame to remove `autoprefixer` config in this case.

ai:
> Furthermore, just because a build doesn't require prefixing in the present doesn't mean that a few weeks in the future wouldn't introduce new properties that need prefixing.

Browsers stopped adding new vendor prefixes. It is very unlikely that we will have any new vendor prefixes in the future.

>  In dev there may well be no targets and the warnings will needlessly show up...

What Browserslist config do you have?

malixsys:
```json
"browserslist": {
    "production": [
      "last 2 iOS major versions",
      "ie 11"
    ],
    "development": [
      "last 1 chrome major versions"
    ]
}
```

ai:
Strange. You should have prefixes for both environments.

Can you debug the code to find why do you have empty prefixes list?

malixsys:
What prefixes should be found for `last 1 chrome major versions`?

ai:
@malixsys you can check it by running `npx autoprefixer --info`

```
➜ BROWSERSLIST="last 1 chrome major versions" npx autoprefixer --info
Browsers:
  Chrome: 81

These browsers account for 8.71% of all users globally

Properties:
  appearance: webkit
  background-clip: webkit
  box-decoration-break: webkit
  color-adjust: webkit
  mask-border-outset: webkit
  mask-border-repeat: webkit
  mask-border-slice: webkit
  mask-border-source: webkit
  mask-border-width: webkit
  mask-border: webkit
  mask-clip: webkit
  mask-composite: webkit
  mask-image: webkit
  mask-origin: webkit
  mask-position: webkit
  mask-repeat: webkit
  mask-size: webkit
  mask: webkit
  text-emphasis-color: webkit
  text-emphasis-position: webkit
  text-emphasis-style: webkit
  text-emphasis: webkit

Values:
  cross-fade: webkit
  fill-available: webkit
  fill: webkit
  image-set: webkit
  stretch: webkit
```

[...]

PR #1409, opened 2021-05-26
Clarify warning about no targets

Pwntheon:
Due to user error or outdated scripts, users may be warned about no prefixes being applied.
Rewrite this error message to be more understandable.

This should help with #1318


ai:
I fixed text a little (we _do not_ show this warning on no browsers). c1b26d1

I will release it when you check my text fixes

Pwntheon:
Looks good.

ai:
Released in 10.2.6.

PR #1412, opened 2021-07-09
Autoprefix the file-selector-button pseudo-element

### LINKED ISSUE #1374: `::file-selector-button` support
It seems [`::file-selector-button`](https://caniuse.com/?search=file-selector-button) isn't automatically prefixed using autoprefixer v10.0.2.

Opening the issue as a feature request to add support for it.

Thanks in advance!

lukewarlow:
Closes #1374 

~~This is prototype implementation using MDN compat data as a fallback because caniuse doesn't have this feature.~~

Edit: This has been changed to hard code support values instead of using the mdn lib due to its size.


package.json
```diff
@@ -29,6 +29,7 @@
     "postcss": "^8.1.0"
   },
   "dependencies": {
+    "@mdn/browser-compat-data": "^3.3.10",
```
ai, line 32: This dependency is 9 MB, 9 times bigger than `caniuse-lite`.

One of idea is to add selected `@mdn/browser-compat-data` data to `caniuse-lite`.

lukewarlow: Attempts to add the data to caniuse have been made - https://github.com/Fyrd/caniuse/pull/5722. They're denied because the data is already upstream in MDN. In cases like this idk what the best idea going forwards is?

ai: `caniuse-lite` is our own re-pack https://github.com/browserslist/caniuse-lite/

Can I Use provides `caniuse-db` package which is too big (12 MB vs 1 MB of our re-pack).

This is why it is easy to add MDN data to `caniuse-lite`. I do not like that `CANIUSE-lite` will have data not from Can I Use, but have no better idea.

data/prefixes.js
```diff
@@ -521,6 +521,25 @@ f(prefixFullscreen, { match: /x(\s#2|$)/ }, browsers =>
...
+    'firefox 82',
+    'opera 75',
+    'safari 14.1',
+    'android 89',
+    'and_chr 89',
+    'op_mob 63',
+    'and_ff 82',
+    'ios_saf 14.5',
+    'samsung 15.0'
+  ]
```
lukewarlow, line 540: I think this is the correct format. This data is taken straight from MDN (samsung internet version is based on a PR by Samsung internet dev rel)

test/cases/file-selector-button.css
```diff
@@ -0,0 +1,3 @@
+::file-selector-button {
```
ai, line 1: You need to call `file-selector-button` test case in `test/autoprefixer.test.ts` (see hacks block there)

lukewarlow: Think I've added it to the correct location

ai:
Do you have Twitter account to mention you in release tweet?

lukewarlow:
https://twitter.com/luke_warlow

ai:
Thanks. Released in 10.3.

PR #1414, opened 2021-07-12
Fix: Term element in content value gets parsed

### LINKED ISSUE #1387: Term 'element' in 'content' value gets parsed
Version: 10.2.3

If I use the word 'element' in a string value for 'content', the word will be prefixed.

```css
    content: "Element 'div' not allowed as child of element 'span' in this context.";
```

```css
  content: "Attribute width not allowed on -moz-element picture at this point.";
  content: "Attribute width not allowed on element picture at this point.";
```

The same output as input would be expected in this case.

```json
  "browserslist": [
    "> 2%",
    "last 2 years",
    "ie 11",
    "not ie_mob > 0",
    "not dead"
  ]
```

```
├── autoprefixer@10.2.3 
└─┬ stylelint@13.9.0
  └── autoprefixer@9.8.6 
```

lukewarlow:
Closes #1387

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -497,6 +497,9 @@ class Processor {
         return true
       }
     }
+    if (node.type === 'decl' && node.prop === 'content') {
+      return true
+    }
 
     return this.disabled(node, result)
   }
--- a/test/autoprefixer.test.ts
+++ b/test/autoprefixer.test.ts
@@ -86,6 +86,15 @@ let clipper = autoprefixer({
 let example = autoprefixer({
   overrideBrowserslist: ['defaults']
 })
+let content = autoprefixer({
+  overrideBrowserslist: [
+    '> 2%',
+    'last 2 years',
+    'ie 11',
+    'not ie_mob > 0',
+    'not dead'
+  ]
+})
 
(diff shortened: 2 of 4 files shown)


lib/processor.js
```diff
@@ -497,6 +497,9 @@ class Processor {
         return true
       }
     }
+    if (node.type === 'decl' && node.prop === 'content') {
```
ai, line 500:
```suggestion
    if (node.prop === 'content') {
```

Only `Declaration` has `prop`.

test/autoprefixer.test.ts
```diff
@@ -370,6 +381,9 @@ it('transition on vendor specific rule', () => {
 it('ignore prefix in vendor at rules', () => {
   check('at-rules')
 })
+it('ignore content property', () => {
+  check('content')
```
ai, line 385:
```suggestion
  let input = read('content')
  let result = postcss([prefixer('scope')]).process(input)
  expect(result.css).toEqual(input)
```

It will allow to avoid the same content in both files.

ai:
Thanks. Released in 10.3.1.

PR #1415, opened 2021-07-12
Add support for mask-position-x and mask-position-y

### LINKED ISSUE #1245: Please add support mask-position-x
It seems that `mask-position` supported but `mask-position-x` aren't

I think it will be better to add all the properties that Chrome DevTools shows:
```
-webkit-mask-box-image-outset
-webkit-mask-box-image-repeat
-webkit-mask-box-image-slice
-webkit-mask-box-image-source
-webkit-mask-box-image-width
-webkit-mask-clip
-webkit-mask-composite
-webkit-mask-image
-webkit-mask-origin
-webkit-mask-position-x
-webkit-mask-position-y
-webkit-mask-repeat-x
-webkit-mask-repeat-y
-webkit-mask-size
```

lukewarlow:
Closes #1245 

Despite this not being mentioned in the CSS Masking spec, I think there's good enough reason to add them.
It's unlikely the final spec won't include them (even if for nothing other than compat reasons) considering there's direct prefixed counterparts and and they're implemented in Firefox already ([demo](https://jsfiddle.net/jbyoLu26/) for testing in Firefox).

### DIFF
--- a/data/prefixes.js
+++ b/data/prefixes.js
@@ -699,6 +699,8 @@ f(prefixCssMasks, browsers => {
     [
       'mask',
       'mask-position',
+      'mask-position-x',
+      'mask-position-y',
       'mask-size',
       'mask-border',
       'mask-border-outset',


ai:
I will think for a few days (anyway, we had too many releases :D)

ai:
Firefox accepts `-webkit-` prefix here as well. We should use just prefixed version.

It is too dangerous to add non-prefixed version to Autoprefixer without spec. It will require a lot of changes if CSSWG changed the name.

I need at least plan to add this property with this name from CSSWG.

PR #1430, opened 2021-10-14
Add support for prefixing :autofill

lukewarlow:
This MR was written to test the output from the MR on caniuse-lite. https://github.com/browserslist/caniuse-lite/pull/86


lukewarlow:
As far as I know the test coverage should be the same as file-selector-button?

PR #1439, opened 2021-12-17
Fix CLI dependency installation instructions

vkentta:
Without `postcss` fails with error:
`Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'postcss' `


ai:
Thanks

PR #1457, opened 2022-04-23
Fix #1446

### LINKED ISSUE #1446: unexpected NaN when autoplacing grid for IE
Given the code:

```css
.test{grid-row-start:span 3; grid-row-end: 5}
```

autoprefixer (online demo) outputs:

```css
/*
* Prefixed by https://autoprefixer.github.io
* PostCSS: v8.3.6,
* Autoprefixer: v10.3.1
* Browsers: since 2005
*/
.test{-ms-grid-row:span 3;grid-row-start:span 3; -ms-grid-row-span: NaN; grid-row-end: 5}
```

It can also be reproduced in the latest version:

```js
const autoprefixer = require('autoprefixer')
const postcss = require('postcss')

const css = `
/* autoprefixer grid: autoplace */
.test{grid-row-start:span 3; grid-row-end: 5}
`;

postcss([ autoprefixer ]).process(css).then(result => {
  result.warnings().forEach(warn => {
    console.warn(warn.toString())
  })
  console.log(result.css)
});
```

https://runkit.com/sukkaw/postcss-issue-1446

SukkaW:
The PR fixed issue #1446.

All existing test cases pass, and a corresponding test case is added.

### DIFF
--- a/lib/hacks/grid-end.js
+++ b/lib/hacks/grid-end.js
@@ -1,4 +1,5 @@
 let Declaration = require('../declaration')
+let { isPureNumber } = require('../utils')
 
 class GridEnd extends Declaration {
   /**
@@ -26,8 +27,12 @@ class GridEnd extends Declaration {
         startDecl = d
       })
       if (startDecl) {
-        let value = Number(decl.value) - Number(startDecl.value) + ''
-        clonedDecl.value = value
+        if (isPureNumber(startDecl.value)) {
+          let value = Number(decl.value) - Number(startDecl.value) + ''
+          clonedDecl.value = value
+        } else {
+          return undefined
+        }
       } else {
         decl.warn(
           result,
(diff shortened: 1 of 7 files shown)


SukkaW:
@ai That's weird, why did only the test on Node.js 17 fail?

Semigradsky:
@SukkaW 
> ERROR: Coverage for lines (99.97%) does not meet global threshold (100%)

Coverage is calculated for Node.js 17 only.
By the way, it would be nice to add Node.js 18 to test worflow.

SukkaW:
> Coverage is calculated for Node.js 17 only.

Let me see what I can do.

> By the way, it would be nice to add Node.js 18 to test workflow.

This should be done in a separate PR.

ai:
Thanks. Released in 10.4.5.

PR #1462, opened 2022-05-24
chore: Included githubactions in the dependabot config

nathannaveen:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+      interval: "weekly"


ai:
I used dependency bot for a few years and didn’t find it really useful.

Instead I:
1. Try to reduce and control my dependencies.
2. Call `pnpm update` in the beginning of any work.

Also Dependabot is useless here because we use `pnpm` https://github.com/dependabot/dependabot-core/issues/1736

PR #1468, opened 2022-07-16
feat: add `dppx` when the browser does not support the `x` unit in resolution media queries

yisibl:
[css-values-4](https://github.com/w3c/csswg-drafts/issues/2631) adds `x` as an alias for `dppx` unit, but not every browser in media queries supports `x` unit. So I think it would be better to add a fallback for `x` unit.

`x` unit support in: `Chrome 68+, Edge 79+, Opera 55+, Safari 16+, Firefox 62+`

See also: 
- https://chromestatus.com/feature/5150549246738432
- https://developer.mozilla.org/en-US/docs/Web/CSS/resolution#browser_compatibility
- Caniuse PR: https://github.com/Fyrd/caniuse/pull/6381

image

### DIFF
--- a/test/cases/resolution.css
+++ b/test/cases/resolution.css
@@ -5,7 +5,7 @@
 
 @media (min-resolution: 144dpi) { }
 
-@media (min-resolution: 2x) { }
+@media (min-resolution: 2dppx), (min-resolution: 2x) { }
 
 @media (min-resolution: 120dpi) { }
 
--- a/test/cases/resolution.css
+++ b/test/cases/resolution.css
@@ -5,6 +5,8 @@
 
 @media (min-resolution: 144dpi) { }
 
+@media (min-resolution: 2x) { }
+
 @media (min-resolution: 2dppx), (min-resolution: 2x) { }
 
 @media (min-resolution: 120dpi) { }
--- a/lib/resolution.js
+++ b/lib/resolution.js
@@ -4,7 +4,7 @@ let Prefixer = require('./prefixer')
 let utils = require('./utils')
 
 const REGEXP = /(min|max)-resolution\s*:\s*\d*\.?\d+(dppx|dpcm|dpi|x)/gi
-const SPLIT = /(min|max)-resolution(\s*:\s*)(\d*\.?\d+)(dppx|dpcm|dpi|x)/i
+const SPLIT = /((min|max)-resolution)(\s*:\s*)(\d*\.?\d+)(dppx|dpcm|dpi|x)/i
 
 class Resolution extends Prefixer {
   /**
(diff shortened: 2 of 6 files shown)


test/cases/resolution.css
```diff
@@ -5,7 +5,7 @@
 
 @media (min-resolution: 144dpi) { }
 
-@media (min-resolution: 2x) { }
+@media (min-resolution: 2dppx), (min-resolution: 2x) { }
```
ai, line 8: Can you also add `@media (min-resolution: 2x) {}` use case?

yisibl: I tried to add it, but it fails in the `removes unnecessary prefixes` test.

I don't know how to handle it better?

ai: Can you add test, so I will see an error on CI?

yisibl: Yeah, https://github.com/postcss/autoprefixer/runs/7377361678?check_suite_focus=true

ai: Hm. Do we need to add `2dppx`? Seems like all browsers which need `2dppx` also works with prefixed values.

yisibl: Do you mean delete the x units and keep only the dppx?

yisibl:
I have added a PR of `x` unit for caniuse as well. https://github.com/Fyrd/caniuse/pull/6381

test/cases/resolution.css
ai: Input:

```css
@media (min-resolution: 2x) { }
```

My suggested output:

```css
@media (-webkit-min-device-pixel-ratio: 2), (-o-min-device-pixel-ratio: 2/1), (min-resolution: 2x) { }
```

yisibl: If the browser is configured with `chrome >= 29`, then only the following will be output:

```
@media (min-resolution: 1x) { }
```

But this syntax is only supported by `Chrome >= 68`.

See test case: https://codepen.io/yisi/pen/rNdjdVp

yisibl: So output `dppx` units at the same time as this PR to prevent browsers from not supporting it.

ai: Seems like these problems are coming because we are mixing prefixes and polyfills here. As a prefixer, Autoprefixer should remove all code which it added.

* What do you think of adding `2x` support to `postcss-present-env` instead?
* Another option is to add `2x` support to Autoprefixer but do not insert `2dppx`. Instead, we can ask users to add polyfill manually (or print a warning after checking `result.processor.plugins`).

yisibl: Well, I think it's okay to print a warning.

ai: We can use this logic:

If:
1. There is `2x` syntax
2. And no `2dppi`
3. And there is specific Chrome versions (we can hardcore versions until Can I Use will merge PR)

We will ask to add `postcss-preset-env` and put it before Autoprefixer

But we may need to create an issue or a plugin for `postcss-preset-env` (you can create this plugin as you did for minmax and ask them to add it to preset).

romainmenke: This would be easy to implement with our new media query parser : https://www.npmjs.com/package/@csstools/media-query-list-parser

Happy to create a plugin for this if work hasn't started yet @yisibl?

Blgwp21 (approved):

PR #1479, opened 2022-09-20
fix: unitless zero for `<angle>` values are allowed in CSS gradients

yisibl:
Problems found during lightningcss discussion: https://github.com/parcel-bundler/lightningcss/issues/298

### DIFF
--- a/test/cases/gradient.out.css
+++ b/test/cases/gradient.out.css
@@ -220,3 +220,37 @@ div {
   background-image: url("https://test.com/lol(test.png"), -o-radial-gradient(yellow, black, yellow);
   background-image: url("https://test.com/lol(test.png"), radial-gradient(yellow, black, yellow);
 }
+
+.unitless-zero {
+  background-image: -webkit-gradient(linear, left bottom, left top, from(green), to(blue));
+  background-image: -webkit-linear-gradient(bottom, green, blue);
+  background-image: -o-linear-gradient(bottom, green, blue);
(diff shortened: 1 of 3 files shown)


test/cases/gradient.out.css
```diff
@@ -220,3 +220,37 @@ div {
...
+  background: -o-repeating-linear-gradient(bottom, blue, red 33.3%);
+  background: repeating-linear-gradient(0, blue, red 33.3%)
+}
+
+.zero-grad {
+  background: -webkit-gradient(linear, left bottom, left top, from(green), to(blue));
+  background: -webkit-linear-gradient(bottom, green, blue);
+  background: -o-linear-gradient(bottom, green, blue);
+  background: linear-gradient(0grad, green, blue);
+  background-image: -webkit-repeating-linear-gradient(bottom, blue, red 33.3%);
```
yisibl, line 239: I don't know why the test failed here, the single test(`test.only('changes angle in gradient'`) was passed.

ai: It was prefixes cleaning test which failed.

Seems like we forgot to think about cleaning repeating gradients when we added their support.

Here is my fix 542bd85

ai:
Thanks. Released in 10.4.12.

test/cases/gradient.out.css
yisibl: Nice

PR #1488, opened 2023-01-26
feat: Show warning message if the user using browserslist config and `overrdeBrowserslist` option

arayaryoma:
This PR introduces a new warning message for users.
When the project using Autoprefixer has .browserslistrc (or `browserslist` field in package.json) and the user specify `overrideBrowserslist` option for autoprefixer, display a waring message to inform the config will be overridden.

This may help users to avoid accidentally overriding .browserslistrc.

## images
image

### DIFF
--- a/lib/autoprefixer.js
+++ b/lib/autoprefixer.js
@@ -54,6 +54,11 @@ function timeCapsule(result, prefixes) {
   )
 }
 
+function findBrowserslistConfigInProject() {
+  let dir = process.cwd();
+  return Boolean(browserslist.loadConfig({ path: dir }))
+}
+
 module.exports = plugin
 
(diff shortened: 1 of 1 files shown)


ai:
What if user need to use Autoprefixer with another options? (For example to clean all prefixes)

arayaryoma:
This warning message will be shown only when using `overrideBrowserslist`, so this change doesn't affect to any other options. ( I'm sorry if my understand was wrong about your concern. )

ai:
To clean some sources from prefixes I will use `overrideBrowserslist` too.

Sorry, this warning will have too many false-positive errors.

arayaryoma:
That makes sense. thanks!

PR #1498, opened 2023-07-24
feat: `::backdrop` using `@mdn/browser-compat-data`

yisibl:
Before that, we added the `-wekbit-` prefix to Safari, which is not really needed.

Waiting for upstream PR: https://github.com/browserslist/caniuse-lite/pull/116

### DIFF
--- a/test/cases/backdrop.out.css
+++ b/test/cases/backdrop.out.css
@@ -1,0 +1,6 @@
+::-ms-backdrop {
+  color: green;
+}
+::backdrop {
+  color: green;
+}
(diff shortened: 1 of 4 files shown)


ai:
I merged `caniuse-lite` changes. Let’s wait for next `caniuse-db` release.

ai:
Thanks. The fix was released at 10.4.15.

test/cases/backdrop.out.css
```diff
@@ -0,0 +1,6 @@
+::-ms-backdrop {
+  color: green;
+}
```
yisibl, line 3: I suspect that IE hasn't implemented this pseudo-element either, which I'm afraid requires further investigation and testing.

PR #1499, opened 2023-08-14
Remove deprecated browsers from README

coliff:
Brackets and Atom code editors have long since been deprecated and neither plugin has been updated in even longer. I think it makes sense to remove links to these.

REF:
- https://github.com/sindresorhus/atom-autoprefixer
- https://github.com/mikaeljorhult/brackets-autoprefixer


ai:
Thanks

PR #1500, opened 2023-09-17
improve performance

romainmenke:
```
Lightning CSS:    5 ms   (7.5 times faster)
Stylis:           6 ms   (6.3 times faster)
Autoprefixer dev: 29 ms  (1.2 times faster)
Autoprefixer:     36 ms
Stylecow:         204 ms (5.7 times slower)
```

- do cheap checks before expensive checks
- loop inside walkers instead of the other way around

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -458,12 +458,12 @@ class Processor {
    * Check for grid/flexbox options.
    */
   disabledDecl(node, result) {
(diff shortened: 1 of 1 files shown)


ai:
Looks amazing. Sorry, I have no time today for PostCSS. I will try to release everything together.

PR #1542, opened 2026-01-30
perf: reduce array allocations

Cherry:
While benchmarking our build on a large project, I spotted a few places where `concat()` in loops could be replaced with `push()` to avoid repeated array allocations.

I imagine most projects won't notice this micro-optimisation, but on stylesheets with many transitions we did see ~15-20% faster processing.

### DIFF
--- a/lib/hacks/gradient.js
+++ b/lib/hacks/gradient.js
@@ -352,7 +352,7 @@ class Gradient extends Value {
 
     node.nodes = []
     for (let param of params) {
-      node.nodes = node.nodes.concat(param)
+      node.nodes.push(...param)
     }
 
     node.nodes.unshift(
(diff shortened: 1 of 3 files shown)


gianpaj:
Hi @Cherry , can you share the benchmark and results? Did you use Node.js or another engine?

I tested this browser benchmark on Chrome 143 and the plain `.push()` was faster, though not by much:
image


https://jsbench.me/cgjfc79bgx/1

Cherry:
Hey @gianpaj, I was testing this against our real code-base so don't have an isolated benchmark to share unfortunately. I was using Windows 11, and Nodejs 24.13.0. I was seeing processing drop from ~55ms to ~45ms on a ~200kb stylesheet.

It's definitely a micro-optimisation, but an easy improvement I saw while benchmarking other stages of our build.

charmander:
@gianpaj *Repeated* `concat` of n arrays of size m one-by-one (like the patched cases in this PR) takes time proportional to n²m. The benchmark doesn’t capture that, since it only concatenates once. (It also doesn’t actually call `push` – note the brackets.)

Issue #1533, opened 2025-03-20
:placeholder-shown is replaced with -moz-placeholder, although this is outdated


kliehm:
Due to a change in autoprefixer in the last two weeks I assume, suddenly `:placeholder-shown` is replaced with `:-moz-placeholder`. According to caniuse, this should only be necessary in Firefox < 50, not in current versions.

Although as far as I understand, `:-moz-placeholder` should be a synonym to `::placeholder`, i.e. to style the placeholder, whereas `:placeholder-shown` is an indicator if the placeholder is visible or not. So if the focus is within the input and the placeholder text disappears, `:placeholder-shown` is false. That's something different than styling the placeholder.

Our `.browserslistrc` looks like this, so Firefox < 50 should not be relevant:
```
> 1% in DE
last 2 versions
not dead
```
For the time being, I disabled replacement of `:placeholder-shown` with `/* autoprefixer: ignore next */`.

ai:
cc @Marukome0743

ai:
> Although as far as I understand, :-moz-placeholder should be a synonym to ::placeholder, i.e. to style the placeholder, whereas :placeholder-shown is an indicator if the placeholder is visible or not

https://caniuse.com/?search=placeholder-shown says that `:-moz-placeholder` equals to `:placeholder-shown`.

Also, another argument is that `:placeholder-shown` and `:-moz-placeholder` both uses single `:` (which means it is a state, not [virtual] element).

But we may be wrong. Please show your proofs or arguments.

kliehm:
> https://caniuse.com/?search=placeholder-shown says that `:-moz-placeholder` equals to `:placeholder-shown`.

Yes, but that's only true for Firefox prior v50. Modern Firefoxes support `:placeholder-shown`.

Here is a minimal example: show the label only if the placeholder is visible:

HTML:
`<input placeholder="&nbsp;" /><label>Text</label>`

CSS:
`label { display: none };`
`input:not(:placeholder-shown) + label { display: block }`

In retrospective, I could have achieved the same effect with `:empty`, but this is how we've done it. It worked until recently. With the replacement it doesn't work as expected anymore in current versions of Firefox:

`input:not(:-moz-placeholder) + label { display: block }`

kliehm:
> Can you show simple example of input CSS, output and expected output?

Here it still works: the search field at the top has a label "what are you looking for" that disappears, when the input receives focus. Also the reset button (the x) only appears when there is text in the input.

[...]

PR #1545, opened 2026-02-27
Remove wrong placeholder-shown prefixing for Firefox

timlg07:
Replacing `:placeholder-shown` with `:-moz-placeholder` breaks in Firefox. While placeholder-shown is working fine (supported since Jan. 2017), the `:-moz-placeholder` added by autoprefixer breaks selectors with `:not`:

Example:

```
input[placeholder]:not(:placeholder-shown) ~ label {
    top: 16px;
    font-size: 12px;
}
```
produces:
```
.form .field .input input[placeholder]:not(:-moz-placeholder)~label,
.form .field .input input[placeholder]:not(:placeholder-shown)~label{
    top: 16px;
    font-size: 12px;
}
```
This always applies, even if the placeholder is shown, because `:not(:-moz-placeholder)` evaluates to true.

I couldn't find any resources stating that `:-moz-placeholder` behaves like `:placeholder-shown`, so I wonder why it's in here?

### DIFF
--- a/lib/hacks/placeholder-shown.js
+++ b/lib/hacks/placeholder-shown.js
@@ -5,9 +5,7 @@ class PlaceholderShown extends Selector {
    * Return different selectors depend on prefix
    */
   prefixed(prefix) {
-    if (prefix === '-moz-') {
-      return ':-moz-placeholder'
-    } else if (prefix === '-ms-') {
+    if (prefix === '-ms-') {
       return ':-ms-input-placeholder'
     }
     return `:${prefix}placeholder-shown`


ai:
Have you just break all `:-moz-placeholder`?

I [know about this problem](https://github.com/postcss/autoprefixer/issues/1533), but we need careful solution.

I would like to accept PR, but I need from you to think about all edge cases.

timlg07:
Do you know that it actually works in Firefox < 50? Because I couldn't find any proof for it. But even if it does, it should only be activated for Firefox < 50, as it's not needed in modern version and even breaks some selectors, like shown in my example🙈

ai:
> Do you know that it actually works in Firefox < 50?

It is mark as supported in Can I Use

https://caniuse.com/?search=placeholder-shown

> it should only be activated for Firefox < 50

We are adding it because of KaiOS 2.5 (FF based browser for cheap phones very popular in Afrika). It is part of default browser set.

timlg07:
That explains why adding `'not Firefox < 51'` to the browserlist had no effect. I adjusted my browserlist and it works now. So I guess this stays and is intended? Thank you for the fast response!

PR #1548, opened 2026-04-06
Add prefixing support for mask-position-x and mask-position-y

### LINKED ISSUE #1537: `mask-position-y` is not prefixed (as `-webkit-mask-position-y`)
`-webkit-mask-position-y` is [supported ](https://caniuse.com/?search=mask-position-y) in just about every browser, but `mask-position-y` is not prefixed.

toporek:
Closes #1537

Adds `-webkit-mask-position-x` and `-webkit-mask-position-y` to the css-masks prefix list. Also adds `-x` support as requested in the issue.

These properties have [97%+ global browser support](https://caniuse.com/mdn-css_properties_-webkit-mask-position-y).

### DIFF
--- a/data/prefixes.js
+++ b/data/prefixes.js
@@ -750,6 +750,8 @@ f(prefixCssMasks, browsers => {
     [
       'mask',
       'mask-position',
+      'mask-position-x',
+      'mask-position-y',
       'mask-size',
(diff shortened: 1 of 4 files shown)


ai:
Thanks.

PR #1550, opened 2026-05-31
fix: loss of browserslist options in `@supports`

charmander:


ai:
Thanks. It is OK if I will release it a little later?

I also want to add Staged Releases.

charmander:
Thank you! And of course, that’s great news. 🙂

PR #1553, opened 2026-08-19
Remove prefixed values overridden by a later declaration

### LINKED ISSUE #1461: `-ms-` vendor prefixes not always removed
If you are using an empty browserlist (or `>100%`) to remove existing vendor prefixes the `-ms-` prefix is not always removed.

The output is as expected given the following CSS:

```css
.example {
    display: -webkit-box;
    display: -ms-flexbox;
    display: flex;
}
```
which produces:

```css
.example {
    display: flex;
}
```

But let's modify the input display property from `flex` -> `none`:

```css
.example {
    display: -webkit-box;
    display: -ms-flexbox;
    display: none;
}
```
The `-ms` prefix is not removed from the output as you'd expect:

```css
.example {
    display: -ms-flexbox;
    display: none;
}
```

mrpmohiburrahman:
Fixes #1461

## Summary

With prefix removal active — an empty browserslist, `>100%`, or simply a prefix that has
become outdated for the configured browsers — `-ms-flexbox` was left behind whenever the
final `display` value was anything other than `flex`:

```css
.example {
  display: -webkit-box;
  display: -ms-flexbox;
  display: none;
}
```

```css
/* before */                /* after */
.example {                  .example {
  display: -ms-flexbox;       display: none;
  display: none;            }
}
```

Worth noting the asymmetry the issue only half-states: `-webkit-box` *was* removed while
`-ms-flexbox` survived, so the output was internally inconsistent rather than merely
incomplete. Changing that last line to `display: flex` made both disappear correctly,
which is what made the bug look value-dependent.

## Root cause

The Values branch of `Processor#remove` decides "is this prefixed declaration one
Autoprefixer generated, or a deliberate hand-written hack?". The repo answers that by
**ordering**: `prefixes.group(decl).down()` walks forward over siblings sharing an
unprefixed property and stops at the first whose value carries no vendor prefix, using
`!Browsers.withPrefix(other.value)` in `lib/prefixes.js`. A prefixed declaration with an
unprefixed sibling below it is a generated fallback and can go; one with nothing below it
is a hack and stays.

The Values callback implemented that question as a raw substring test on the value, and it
was wrong in both directions:

- `'none'.includes('flex')` is `false`, so `-ms-flexbox` was judged a hack even though
  `display: none` sits directly below it and overrides it in every browser. The reported
  symptom.
- `'-ms-flexbox'.includes('flex')` is `true`, so `-webkit-box` accepted the *prefixed*
  `-ms-flexbox` as its "unprefixed sibling" and was removed. That is the only reason the
  two prefixes behaved differently.

The Properties branch just above is unaffected — it tests the property
(`normalize(other.prop) === unprefixed`) rather than the value.

## Change

The callback now asks the question the sibling walker three lines away already asks:

```js
return !Browsers.withPrefix(other.value)
```

Ordering stays the only hack signal: `down()` is still the only direction consulted, so a
prefixed declaration placed *after* the unprefixed one (`test/cases/vendor-hack.css` `.c`)
and a lone prefixed declaration with no sibling below it (`value-hack.css` `.hack`) are
both still preserved. The `unprefixed = checker.unprefixed` assignment that the change
leaves unread is removed with it.

## Intentional behaviour change

Removal now applies whenever a later declaration overrides the prefixed one, even when it
does not name the same feature:

```css
a {
  background: -webkit-linear-gradient(top, white, black);
  background: red;
}
/* before: both kept.   after: `background: red` only. */
```

This is the same reasoning that makes `display: none` remove `-ms-flexbox` — a later
declaration the browser understands always overrides the earlier prefixed one, so the
prefixed declaration is dead code in every browser. It applies under real browserslist
configs, not only empty ones. No existing test covered it, so I would rather name it here
than leave it to be found in review.

Prefixes that are still needed are not affected, since the loop only runs on values
already in the `remove` set. I checked the cases that would matter: with `ie 10` an
`-ms-flexbox` / `flex` pair is left intact, and with `safari 6` a prefixed gradient above
its unprefixed twin is left intact.

## Tests

`test/cases/value-hack.css` gains a `.not-hack3` case covering the reported input. I
reused that fixture rather than adding one — it is already wired to `cleaner` and is
deliberately absent from `COMMONS`, so it is not round-tripped by
`removes unnecessary prefixes`. Reverting `lib/processor.js` while keeping the fixture
turns the case red, so it does pin the fix rather than passing either way.

```
pnpm run unit           246 / 246   (246 / 246 before the change)
pnpm run test:lint      clean
pnpm run test:coverage  100% lines
pnpm run test:size      69.87 kB / 70 kB limit
```

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -682,7 +682,10 @@ class Processor {
         if (!checker.check(decl.value)) continue
 
         let notHack = this.prefixes.group(decl).down(other => {
-          return !Browsers.withPrefix(other.value)
+          if (this.prefixes.browsers.selected.length === 0) {
+            return !Browsers.withPrefix(other.value)
+          }
+          return other.value.includes(checker.unprefixed)
         })
 
         if (notHack) {
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -1,5 +1,6 @@
 let parser = require('postcss-value-parser')
 
+let Browsers = require('./browsers')
 let Value = require('./value')
 let insertAreas = require('./hacks/grid-utils').insertAreas
 
@@ -680,9 +681,8 @@ class Processor {
         if (!checker.check) continue
         if (!checker.check(decl.value)) continue
 
-        unprefixed = checker.unprefixed
         let notHack = this.prefixes.group(decl).down(other => {
-          return other.value.includes(unprefixed)
+          return !Browsers.withPrefix(other.value)
         })
 
         if (notHack) {
--- a/test/autoprefixer.test.js
+++ b/test/autoprefixer.test.js
@@ -25,6 +25,9 @@ let filterer = autoprefixer({
 let borderer = autoprefixer({
   overrideBrowserslist: ['Safari 4', 'Firefox 3.6']
 })
+let fallbacker = autoprefixer({
+  overrideBrowserslist: ['Safari 12']
+})
 let cascader = autoprefixer({
   cascade: true,
   overrideBrowserslist: ['Chrome > 19', 'Firefox 21', 'IE 10']
@@ -169,6 +172,8 @@ function prefixer(name) {
     return keyframer
   } else if (name === 'border-radius') {
     return borderer
+  } else if (name === 'value-fallback') {
+    return fallbacker
   } else if (name === 'gradient') {
     return gradienter
   } else if (name === 'gradient-fix') {
@@ -370,6 +375,10 @@ test('keeps vendor-specific hacks', () => {
 
 test('keeps values with vendor hacks', () => {
   check('value-hack')
+})
+
+test('keeps prefixed value when the next value is not supported', () => {
+  check('value-fallback')
 })
 
 test('works with comments', () => {
--- a/test/cases/value-fallback.css
+++ b/test/cases/value-fallback.css
@@ -1,0 +1,9 @@
+.fallback {
+  width: -webkit-calc(100% - 20px);
+  width: clamp(10rem, 50%, 60rem);
+}
+
+.overridden {
+  display: -webkit-box;
+  display: flex;
+}
--- a/test/cases/value-fallback.out.css
+++ b/test/cases/value-fallback.out.css
@@ -1,0 +1,8 @@
+.fallback {
+  width: -webkit-calc(100% - 20px);
+  width: clamp(10rem, 50%, 60rem);
+}
+
+.overridden {
+  display: flex;
+}
--- a/test/cases/value-hack.css
+++ b/test/cases/value-hack.css
@@ -11,6 +11,12 @@
   transition: filter 1s;
 }
 
+.not-hack3 {
+  display: -webkit-box;
+  display: -ms-flexbox;
+  display: none;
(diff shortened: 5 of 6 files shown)


ai:
```
a {
  background: -webkit-linear-gradient(top, white, black);
  background: red;
}
/* before: both kept.   after: `background: red` only. */
```

No, we can afford this change.

mrpmohiburrahman:
@ai before you merge — this deletes fallbacks that browsers **in** the browserslist still need.

```css
/* ios_saf >= 15 */
height: 100vh;
height: -webkit-fill-available;
height: 100dvh;
```

`-webkit-fill-available` gets deleted, so iOS 15.0–15.3 falls back to the buggy `100vh` (`dvh` is 15.4+). Same with `safari 15` + `color-mix()` and `safari 12` + `clamp()`.

Membership in `remove` only proves no selected browser needs the prefix *for that feature*. It says nothing about a later, differently-named value — `!Browsers.withPrefix(other.value)` drops that link.

Pushed a narrowing to the empty-browserslist case #1461 actually reports:

```js
if (this.prefixes.browsers.selected.length === 0) {
  return !Browsers.withPrefix(other.value)
}
return other.value.includes(checker.unprefixed)
```

Issue still fixed, every other browserslist byte-identical to `main`, new fixture red without it. 247/247.

This walks back the change you approved for non-empty lists. Say so and I'll restore the wide version.

mrpmohiburrahman:
@ai one thing to check before you merge. This change removes fallback lines that browsers in the browserslist still need.

```css
/* ios_saf >= 15 */
height: 100vh;
height: -webkit-fill-available;
height: 100dvh;
```

The middle line now gets deleted. iOS 15.0 to 15.3 has no `dvh` (it landed in 15.4), so those versions fall back to `100vh` and hit the exact bug the snippet exists to avoid. Same story with `safari 15` and `color-mix()`, and with `safari 12` and `clamp()`.

A value being in the `remove` set only tells us that no selected browser needs the prefix for that feature. It tells us nothing about whether those browsers can read the next line. `other.value.includes(unprefixed)` was what connected the two, and `!Browsers.withPrefix(other.value)` dropped that connection.

I pushed a smaller version. #1461 is about an empty browserslist (or `>100%`), which is the only case where removing the line cannot break anyone:

```js
if (this.prefixes.browsers.selected.length === 0) {
  return !Browsers.withPrefix(other.value)
}
return other.value.includes(checker.unprefixed)
```

The reported bug is still fixed. Every other browserslist now gives the same output as `main`. I added `test/cases/value-fallback.css`, which fails without the fix. All 247 tests pass.

This does undo the behaviour change you approved, for non-empty browserslists. If you would rather keep the wide version, tell me and I will put it back.

ai:
I need a full research why it is not breaking changes (PostCSS has too big users base).

And I need you to no copy-paste text from LLM and add some human ownership to this task. Can you do research without any LLM and then use LLM only to review it?

Right now, this PR looks like too dangerous. We are fixing an issue which bother small number of users with risks of breaking something big.

Issue #1505, opened 2024-01-09
Avoid generating prefixes for browsers that don't support Gradient Midpoints


yisibl:
- https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient#color-hint
![image](https://github.com/postcss/autoprefixer/assets/2784308/33166639-47d9-4018-9b48-fba0578b5dbe)
- Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=420527

```css
/*
* Prefixed by https://autoprefixer.github.io
* PostCSS: v8.4.14,
* Autoprefixer: v10.4.7
* Browsers: > 0%
*/

.example {
    background: -webkit-gradient(linear, left top, left bottom, color-stop(10%, red), color-stop(30%), color-stop(90%, blue)); /* invalid */
    background: -webkit-linear-gradient(red 10%, 30%, blue 90%); /* invalid */
    background: -moz-linear-gradient(red 10%, 30%, blue 90%);
    background: -o-linear-gradient(red 10%, 30%, blue 90%); /* invalid */
    background: linear-gradient(red 10%, 30%, blue 90%);
}
```

- I'm not sure if -webkit-gradient supports this syntax, need to investigate further.
- Interestingly, Firefox supports `-webkit-linear-gradient(red 10%, 30%, blue 90%)`

ai:
I will need PR for that optimization

PR #1554, opened 2026-09-09
Fix old WebKit gradients with color hints

ryanchou1994:
With Browserslist set to `Chrome 25, Opera 12, Android 2.3`, `linear-gradient(red 10%, 30%, blue 90%)` currently produces an old WebKit gradient containing `color-stop(30%)`. That syntax needs both a position and a color, so it cannot represent a color hint.

Skip the old `-webkit-gradient` conversion when a standalone hint is present. The standard gradient and the other prefixed forms are preserved. The regression tests cover numeric and calculated hints, comments, directions, ordinary color stops, and multiple background layers.

Related to #1505. This addresses the invalid old WebKit conversion; it does not change which other vendor forms are generated.

Validation: the new regression tests fail before the fix; `pnpm test` passes on macOS and Linux (250 tests, 100% line coverage, lint and size checks). Historical browser rendering was not tested.

### DIFF
--- a/lib/hacks/gradient.js
+++ b/lib/hacks/gradient.js
@@ -361,6 +361,17 @@ class Gradient extends Value {
     }
 
     this.oldDirection(params)
+    for (let param of params.slice(1)) {
+      let tokens = param.filter(i => i.type === 'word' || i.type === 'function')
+      if (
+        tokens.length === 1 &&
+        ((tokens[0].type === 'word' && parser.unit(tokens[0].value)) ||
+          (tokens[0].type === 'function' &&
+            /^(calc|min|max|clamp)$/i.test(tokens[0].value)))
+      ) {
+        return false
+      }
+    }
     this.colorStops(params)
 
     node.nodes = []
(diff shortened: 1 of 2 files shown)


ai:
Can you help me a little with review?

Can you show CSS input, current CSS output and PR's output?

ryanchou1994:
Sure. With `overrideBrowserslist: ['Android 2.3', 'Chrome 25']`:

Input:

```css
a {
  background: linear-gradient(red 10%, 30%, blue 90%);
}
```

Current output:

```css
a {
  background: -webkit-gradient(linear, left top, left bottom, color-stop(10%, red), color-stop(30%), color-stop(90%, blue));
  background: -webkit-linear-gradient(red 10%, 30%, blue 90%);
  background: linear-gradient(red 10%, 30%, blue 90%);
}
```

With this PR:

```css
a {
  background: -webkit-linear-gradient(red 10%, 30%, blue 90%);
  background: linear-gradient(red 10%, 30%, blue 90%);
}
```

The standalone `30%` is a color hint, but the current conversion turns it into `color-stop(30%)`, which has no color argument. The PR omits that invalid old-WebKit declaration and preserves the other two declarations.

ai:
I added a warning also so developer will know that the syntax would not be supported by all browsers da65c36

ai:
Thanks. Released in 10.5.6.

Issue #1243, opened 2019-09-09
Autoprefixer computing row and column numbers incorrectly


silverAndroid:
Possibly the same as #1181 and #1241

In IE 11, even though each item was given the right row & column by default, when the `.change` class added, the row and columns ignore the gap given to the .container class style and assign b & c to the gaps.

Codepen: https://codepen.io/silverAndroid/pen/RwbMPpe

ai:
Are you talking about CSS Grid? @Dan503 could help you here.

Also, check https://github.com/postcss/autoprefixer/issues/1237

Dan503:
Nope, this is a unique issue. @ai can you re-open?

https://github.com/postcss/autoprefixer/issues/1181 = an issue with Autoprefixer handling number based grid-area declarations

https://github.com/postcss/autoprefixer/issues/1241 = An old bug that we already fixed ages ago

https://github.com/postcss/autoprefixer/issues/1237 = to do with autoplacement and taking `display:none` elements into account.

This issue has to do with the grid gap not inheriting when changing the grid template.

## Input CSS

```css
.container {
  display: grid;
  grid-column-gap: 50px;
  grid-row-gap: 50px;
  grid-template-columns: repeat(2, 1fr);
  grid-template-areas: 
    "a b"
    "c .";
}

.container.change {
  grid-template-areas: 
      "a b"
      "c .";
}

.a {
  grid-area: a;
}

.b {
  grid-area: b;
}

.c {
  grid-area: c;
}
```

## Current output CSS

```css
.container {
  display: -ms-grid;
  display: grid;
  grid-column-gap: 50px;
  grid-row-gap: 50px;
  -ms-grid-columns: 1fr 50px 1fr;
  grid-template-columns: repeat(2, 1fr);
  -ms-grid-rows: auto 50px auto;
      grid-template-areas: 
    "a b"
    "c .";
}

.container.change {
      grid-template-areas: 
      "a b"
      "c .";
}

.a {
  -ms-grid-row: 1;
  -ms-grid-column: 1;
  grid-area: a;
}

.container.change > .a {
  -ms-grid-row: 1;
  -ms-grid-column: 1;
}

.b {
  -ms-grid-row: 1;
  -ms-grid-column: 3;
  grid-area: b;
}

.container.change > .b {
  -ms-grid-row: 1;
  -ms-grid-column: 2;
}

.c {
  -ms-grid-row: 3;
  -ms-grid-column: 1;
  grid-area: c;
}

[...]

Issue #1181, opened 2019-01-18
[css-grid] Autoprefixer outputting wrong IE grid-span syntax


cloudsociety:
As mentioned in #1064, I'm seeing similar behaviour with `span x` not prefixing correctly with `grid-gap`. 

When declaring `grid-area: 1 / 2 / 1 / span 4;`, it is being prefixed to the following: 

```
    -ms-grid-row: 1;
    -ms-grid-column: 2;
    -ms-grid-column-span: 4;
    grid-area: 1 / 2 / 1 / span 4;
```

But `-ms-grid-column-span: 4;` should actually be `-ms-grid-column-span: 7;` to account for the grid gaps.

Test case: https://codepen.io/davecross/pen/ce7e08421ea29a8464ff82b84fcc7d12?editors=1100

Dan503:
Here is a cleaner reduced test case of the problem:

## Input CSS

```css
.grid {
  display: grid;
}

.grid--2col {
  grid-gap: 20px;
  grid-template-columns: repeat(12, 1fr);
  grid-template-areas: "a a a a a a b b b b b b";
}

.grid--2col > .a {
  grid-area: a;
}

.grid--2col > .b {
  grid-area: b;
}

@media (min-width: 376px) {
  .grid--2col .smaller {
    grid-area: 1 / 2 / 1 / span 4;
  }
}
```

## Current output CSS

```css
.grid {
  display: -ms-grid;
  display: grid;
}

.grid--2col {
  grid-gap: 20px;
  -ms-grid-columns: 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr 20px 1fr;
  grid-template-columns: repeat(12, 1fr);
      grid-template-areas: "a a a a a a b b b b b b";
}

.grid--2col > .a {
  -ms-grid-row: 1;
  -ms-grid-column: 1;
  -ms-grid-column-span: 11;
  grid-area: a;
}

.grid--2col > .b {
  -ms-grid-row: 1;
  -ms-grid-column: 13;
  -ms-grid-column-span: 11;
  grid-area: b;
}

@media (min-width: 376px) {
  .grid--2col .smaller {
    -ms-grid-row: 1;
    -ms-grid-column: 2;
    -ms-grid-column-span: 4;
    grid-area: 1 / 2 / 1 / span 4;
  }
}
```

## expected output

```css
.grid {
[...]

PR #1557, opened 2026-09-13
Inherit grid gap set with the row-gap and column-gap longhands

dualfroz:
When a rule inherits `grid-template-areas`, Autoprefixer copied the container's gap only from the `grid-gap` shorthand, so a gap set with `grid-row-gap`/`grid-column-gap` was lost (`-ms-grid-columns: (1fr)[2]` instead of `1fr 50px 1fr`).
Both forms now go through the regex `getGridGap` already used. This may be behind #1243 and #1181, but I could not confirm which syntax was used there.

Refs #1243


ai:
Released in 10.6.1.

